PR #6, opened 2019-01-30
Add latex syntax

zlliang:


octref:
Thanks! Just wondering, did you take this from https://github.com/textmate/latex.tmbundle?
I'd like to keep a table of grammars and their upstreams.

zlliang:
In fact I take [this](https://github.com/James-Yu/LaTeX-Workshop/blob/master/syntax/LaTeX.plist), which is still being maintained.

PR #11, opened 2019-04-04
Use json5 for full theme compat

wesbos:
VS Code uses JSON5 to parse themes, which is why I can have comments in my theme file. For Shiki to work with all themes, this package should use JSON5 too . Thanks!


octref:
Looks good. Thanks!

PR #28, opened 2020-02-11
escapeHtml bug fix

jackens:
Conf. issue <https://github.com/octref/shiki/issues/26>


vfcp:
`"` characters should also be escaped to `&quot;` and `'` should be escaped to `&#39;`

See https://github.com/component/escape-html/blob/b42947eefa79efff01b3fe988c4c7e7b051ec8d8/index.js#L33 as reference

octref:
@vfcp thanks for pointing it out. I'll clean it up after merging.

PR #37, opened 2020-04-29
(java): added 'strictfp' to storage modifiers

### LINKED ISSUE #36: Java - strictfp modifier
From the Java Language Specification:

```
8.1.1.3 strictfp Classes
The effect of the strictfp modifier is to make all float or double expressions
within  the  class  declaration  (including  within  variable  initializers,  instance
initializers, static initializers, and constructors) be explicitly FP-strict (§15.4).
This implies that all methods declared in the class, and all nested types declared in
the class, are implicitly strictfp.
```

This modifier doesn't exist in the `java.json` file. So the syntax highlighting can't work for that.

0xflotus:
* fixes #36

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### 0.2.0
+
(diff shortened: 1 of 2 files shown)


octref:
I retrieved latest java tmLanguage from https://github.com/microsoft/vscode/blob/master/extensions/java/syntaxes/java.tmLanguage.json and it includes `strictfp`. Still, thanks a lot 🙏

PR #48, opened 2020-07-18
feat: add clarity language

aulneau:
This adds the clarity language. I tried to yarn link the packages but kept getting an error that the language was not registered. Are there any more steps for adding a language @octref?

Thanks! :)


octref:
We have added an API to allow registering custom languages: #49.
We are only adding top-50 languages in TIOBE for now to keep the package small, please see #55.

Issue #42, opened 2020-06-08
Twig, PHP syntax highlighting?


thamas:
Thanks for this great tool!

I use Shiki in Gridsome via https://gridsome.org/plugins/gridsome-plugin-remark-shiki

I wonder how it is possible to add syntax highlighting for Twig and PHP code snippets which seems to be missing currently.

octref:
We already bundle PHP, but not twig. I'll think a bit about adding twig language. You can also wait for #49 to land.

D-Marc1:
@octref PHP isn't working for me either, but JavaScript and CSS seem fine.

PR #47, opened 2020-07-18
add arc lang

GalAster:
Used to enhance the highlight of web pages

- highlight effect in vsc: [vscode-arc](https://github.com/GalAster/vscode-arc)
- highlight effect in web: [migrate from json](https://arc-lang.netlify.app/cn/migrate/From%20JSON.html)

PR #49, opened 2020-07-25
Adds the ability to have custom tmlanguages from outside shiki

### LINKED ISSUE #10: How to use tmLanguage files by path?
How do I use a custom grammar, in the form of a string of it's path for the syntax highlighting in shiki?
I want to do this in the function `shiki.codeToHtml()`.

### LINKED ISSUE #44: Dart Language Support
Couldn't find dart in shiki languages list. Is it there but just not listed ?

orta:
Fixes #10 

IMO, I think you can close -  #44  #48 #47 too because then they can use their own themes.

Updates docs too!

/cc @pngwn


GalAster:
nice work, also add interface here:

https://github.com/octref/shiki/blob/43e03b8b506b13bf93b151b684919560d7b38598/packages/vuepress-plugin/index.js#L7-L10

orta:
It's an optional field, it doesn't need to be added there 👍

hacknug:
I think this would also close #42, no?

packages/languages/src/lang.ts
```diff
@@ -111,7 +114,8 @@ export function getLangRegistrations(langs: TLang[]): ILanguageRegistration[] {
     })
   })
 
-  return langs.map(l => langRegistrationMap[l])
+  const shikiLanguages = langByName.map(l => langRegistrationMap[l])
+  return [...langReferences, ...shikiLanguages]
```
octref, line 118: I think something like below would be clearer in intent:

```ts
export function getLangRegistrations(langs: (TLang | ILanguageRegistration)[]): ILanguageRegistration[] {
  const langByName = langs.filter(l => typeof l === "string") as string[]
  const langReferences = langs.filter(l => typeof l !== "string")

  const langRegistrationMap = {}
  languages.forEach(l => {
    langRegistrationMap[l.id] = l
    l.aliases.forEach(a => {
      langRegistrationMap[a] = l
    })
  })
  const shikiLanguages = langByName.map(l => langRegistrationMap[l])

  return [...langReferences, ...shikiLanguages]
}
```

Also, have you thought about conflicts? What if someone registers `ts` again?
Currently with `[...langReferences, ...shikiLanguages]`, duplicated registrations would be ignored, see https://github.com/octref/shiki/blob/43e03b8b506b13bf93b151b684919560d7b38598/packages/shiki/src/highlighter.ts#L69-L75

packages/site/gen-custom-theme.js
```diff
@@ -0,0 +1,28 @@
...
+    highlight: (code, lang) => {
+      return highlighter.codeToHtml(code, lang)
+    }
+  })
+
+  const result = md.render(fs.readFileSync('svelte.md', 'utf-8'))
+  fs.writeFileSync('svelte.html', result)
+
+  console.log('done')
+})
```
octref, line 28: I know you are a big fan of svelte, but considering its popularity we can just add it.
Would you be against choosing a more niche language to illustrate the idea that "if your language isn't popular, you can still colorize it with a custom grammar"? (I can do this part)

packages/site/index.md
```diff
@@ -157,6 +157,28 @@ shiki.getHighlighter({
...
+    path: './svelte.tmLanguage.json',
+    aliases: []
+  },
+  ...shiki.commonLangIds,
+  ...shiki.commonLangAliases
+  ]
+})
+```
+
+<div id="svelte"></div>
```
octref, line 179: You figured this out 😉

packages/site/svelte.md
```diff
@@ -0,0 +1,50 @@
...
+	<p>Please select at least one flavour</p>
+{:else if flavours.length > scoops}
+	<p>Can't order more flavours than scoops!</p>
+{:else}
+	<p>
+		You ordered {scoops} {scoops === 1 ? 'scoop' : 'scoops'}
+		of {join(flavours)}
+	</p>
+{/if}
+```
```
octref, line 50: I'd say use space instead of tabs, and only include half-a-screenful of content to illustrate the point.

packages/site/gen-custom-theme.js
10thfloor: We're looking at a custom integration for [Cadence](docs.onflow.org)

packages/languages/src/lang.ts
orta: Aye, agreed, and people should be able to shoot themselves in the foot if they want to overwrite the existing themes

packages/site/gen-custom-theme.js
orta: hah, Svelte was just what me and @pngwn were working on at the time, I'm game for something more obscure, doing it now 👍

packages/site/svelte.md
orta: Hah, rockstar code doesn't need indenting at all!

orta:
Cool, all feedback addressed

octref:
@orta & @pngwn thanks again for your work 🙌

orta:
first CI and now tests, looks like Shiki is really growing up 🥇

pngwn:
I thought we established that the rendered HTML was a set of acceptance tests in disguise. 

Shiki always had tests!

octref:
🖕 this, master auto publishes to netlify, aren't you pressured to not break it 😝

PR #50, opened 2020-07-29
[languages] Move path.resolve() for better static analysis

styfle:
This moves the `path.resolve()` in place so that static analysis tools such as webpack and node-file-trace can determine that the assets in `../data/grammars` should be bundled.

Fixes https://twitter.com/browniefed/status/1288502416732614656


octref:
Thanks for keeping an eye on the twitter-bug-tracker 👀 
Glad you made webpack work for you.

PR #51, opened 2020-08-02
perf: Avoid some unnecessary string copies

Gerrit0:
TypeDoc@future is using Shiki now for highlighting, and while profiling I noticed that >50% of the runtime in a moderately sized project was spent highlighting code comments. ~60% of this time was spent in node internals, which I deduced was the `substring` calls here.

Unfortunately, due to the existing API contract requiring that the same string be included several times, it isn't possible to completely eliminate this overhead... but I was able to reduce the runtime by ~20% by making these changes. There is still lots of room for improvement. I think the runtime can be cut in half if less information is returned (I don't need all that this gives... maybe I should just be using vscode-textmate directly)


orta:
👋🏻 if you can give this a rebase (so that CI runs) - I'm good to merge this, thanks!

Gerrit0:
@orta rebased as requested.

orta:
rock, thanks

PR #52, opened 2020-08-03
Add an option to skip generating the explanation text

Gerrit0:
As alluded to in #51, there are still lots of string copies that aren't necessarily required when tokenizing due to including the explanation for each token's color. This adds an option which can be used to disable the explanation generation if not required.

Since this builds on the changes in #51, it is also included in this PR.


octref (approved):
Good work! I didn't bother to explain the API earlier since it's simple and mostly documented on the website, but now I think it's better to add some explanations.

PR #53, opened 2020-08-03
Adds CI

orta:
Runs TypeScript across the entire thing, then runs the website generator as it's basically a high-level integration test at this point.


orta:
( It won't run on my PR, because it has to be on master to start taking effect )

PR #64, opened 2020-08-12
fix: fix brace color in dark_plus

### LINKED ISSUE #45: dark_plus theme colors braces black
Not much else to say... This also seems to happen with dark_vs. I spent a bit of time digging into it, but haven't figured out the fix.

image

Enter-tainer:
closes #45

### DIFF
--- a/packages/themes/data/vscode/dark_vs.json
+++ b/packages/themes/data/vscode/dark_vs.json
@@ -4,6 +4,12 @@
 	"include": "./dark_defaults.json",
 	"tokenColors": [
 		{
+			"scope": "source",
+			"settings": {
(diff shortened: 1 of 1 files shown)


packages/themes/data/vscode/dark_vs.json
```diff
@@ -3,6 +3,12 @@
 	"name": "Dark (Visual Studio)",
 	"include": "./dark_defaults.json",
 	"tokenColors": [
+		{
+			"scope": "source",
+			"settings": {
+				"foreground": "#FFFFFF",
```
Enter-tainer, line 9:
```suggestion
				"foreground": "#CCCCCC",
```

PR #65, opened 2020-08-16
Adds a deploy  workflow

orta:
This sets up GitHub to deploy packages when they change on a nightly basis. Sharing the infra with https://github.com/sveltejs/language-tools/ 

So for example, because I merged #51 today - then that package is the only one which would be deployed tomorrow morning

I don't have access to deploy the module, so I'm afraid adding the NPM access token in [the security section](https://github.com/shikijs/shiki/settings/secrets) - my recommendation is to make a shiki-only bot account to deploy so that your credentials don't go in there

### DIFF
--- a/.github/workflows/Deploy.yml
+++ b/.github/workflows/Deploy.yml
@@ -1,0 +1,27 @@
+name: Deploy Daily Builds
+
+# For testing
+# on: push
+
+# For production
+on:
+  schedule:
+    - cron: '0 4 * * *'
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          registry-url: 'https://registry.npmjs.org'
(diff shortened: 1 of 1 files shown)


octref:
I'm thinking if we should publish nightly build behind a tag. We can make it easy to update `vls` dependencies in Vetur, but IMO there should be two modes:

1. For people who want the latest changes/bugfixes
2. For people who need a stable enviroment to get things done

orta:
I do that in svelte too, but both of those reps get a lot of PRs - this repo generally doesn't

octref:
I'll set it up after next release. Doing a lot of changes.

octref:
Did you create https://www.npmjs.com/org/shiki-bot?

octref:
Feeling this would have unintended consequences — for example, I might have half-done changes on master that I want to polish before publishing.
So opting not to use it. I'll add you on NPM for the packages.

orta:
Cool, totally reasonable!

PR #66, opened 2020-08-16
fix: escapeHtml

### LINKED ISSUE #63: Wrong HTML escape sequences in demo
image

It seems that the implementation of html escape is wrong.

``` ts
function escapeHtml(html: string) {
  return html
    .replace(/&/g, '&amp;')
    .replace(/</g, '&lt;')
    .replace(/>/g, '&gt;')
    .replace(/&/g, '&amp;')
    .replace(/'/g, '&apos;')
}
```

It firstly converts `>` to `&gt`, then converts `&gt` to `&ampgt`, which is not what we expected.

 I suggest checking out the implementation of [lodash](https://github.com/lodash/lodash/blob/master/escape.js).

antfu:
closes #63, thanks to @Enter-tainer 's suggestions.

- Remove double escaping for `&`
- Makes the process done in one pass. 
- Replaced `&apos;` with `&#39;`
  - https://stackoverflow.com/questions/2083754/why-shouldnt-apos-be-used-to-escape-single-quotes
- Escaping for double quotes
  - http://wonko.com/post/html-escaping
- The escaping problem in the demo site is fixed as well


orta:
👍🏻 cool, looks good to me!

octref:
I pointed netlify to shikijs/shiki instead of octref/shiki. Should be fixed on demo page now. Thanks for PR 👍

Issue #74, opened 2020-08-20
Highlight support for add/delete/focus lines


codechips:
Not sure if it's possible to highlight lines in Shiki. Haven't found any info in the code and repo. However, while searching for the solution, I accidentally saw that @dceddia already did all the hard work. @dceddia can you please a PR for this? :pray: 

https://github.com/dceddia/shiki-core

dceddia:
Indeed! I've been meaning to make a PR for this and a few other things I changed/fixed - although it looks like a few have already been fixed by other folks! (#66 to fix `&` encoding, #51 & #52 for performance). So I'll have to merge those with mine and make sure nothing broke.

FYI the /shiki-core repo I have (along with shiki-languages and shiki-themes) are just so that I can install these packages in package.json without having to publish them to NPM. I couldn't find a way around the Lerna setup that would allow Netlify to install everything, so I split them up. The source of the changes are (now) in https://github.com/dceddia/shiki

octref:
One thing I'm not sure is whether the metadata for "which line should be highlighted" should live along the source (such as  Markdown) or in code. There are other non-markdown cases which I want to support as well. It seems impossible to find a way to mark highlighted lines in the source for all languages.

Here's vuepress's handling of MD: https://v1.vuepress.vuejs.org/guide/markdown.html#line-highlighting-in-code-blocks

dceddia:
Yeah, where to put the metadata was one of the tricky things to figure out. I handed that responsibility to my own blog code, which parses the code blocks and looks for comments that say which lines to highlight ([see this gist](https://gist.github.com/dceddia/d1d81e832989b3362a17a72b35184ddf)) (my blog uses Jekyll, and integrating it with Shiki took some hackery with sending code over pipes... so that's not all of the code, just the part that parses out the line numbers)

The changes I made to Shiki allow it to accept arrays of line numbers to highlight, so it's sort of format-agnostic in that sense -- it's up to the caller of `codeToHtml` to figure out how to parse out which lines to highlight, and Shiki is just given the plain unannotated code. Here's [an example of it in action](https://daveceddia.com/react-confirmation-modal-state-machine/) and here's one of the code blocks from the markdown:

```js
// focusLines: ['7-20']
// deleteLines: [9]
// addLines: [8]
const confirmationFlow = createMachine(
  /* ... */
  confirming: state(
    transition('confirm', 'loading'),
    transition('cancel', 'initial')
  ),
  loading: invoke(
[...]

PR #77, opened 2020-08-22
feat: wrapLines option for renderToHtml

codepunkt:
fixes #76

PR #83, opened 2020-08-26
Highlighting lines

dceddia:
I'm not sure this is ready to merge, but I wanted to put it up to help the discussion in #74.

The broad idea here is to pass an array of line numbers and ranges into Shiki when calling `codeToHtml`, and it will then add appropriate classes to the highlighted/added/deleted/focused lines.

### DIFF
--- a/packages/shiki/src/__tests__/simple.test.ts
+++ b/packages/shiki/src/__tests__/simple.test.ts
@@ -7,3 +7,60 @@ test('Nord highlighter highlights simple JavaScript', async () => {
   const out = highlighter.codeToHtml(`console.log('shiki');`, 'js')
   expect(out).toMatchSnapshot()
 })
+
+test('Individual lines are highlighted', async () => {
+  const highlighter = await getHighlighter({
+    theme: 'nord'
+  })
+  const code = `
+let a = 1;
+let b = 2;
+console.log(a+b);
+console.log('shiki');
+`
+
+  const out = highlighter.codeToHtml(code, 'js', { highlightLines: [2, 4] })
+  expect(out).toBe(
+    `<pre class=\"shiki highlighted\" style=\"background-color: #2e3440\"><code>
+<span class=\"hl\"><span style=\"color: #81A1C1\">let</span><span style=\"color: #D8DEE9FF\"> </span><span style=\"color: #D8DEE9\">a</span><span style=\"color: #D8DEE9FF\"> </span><span style=\"color: #81A1C1\">=</span><span style=\"color: #D8DEE9FF\"> </span><span style=\"color: #B48EAD\">1</span><span style=\"color: #81A1C1\">;</span></span>
+<span style=\"color: #81A1C1\">let</span><span style=\"color: #D8DEE9FF\"> </span><span style=\"color: #D8DEE9\">b</span><span style=\"color: #D8DEE9FF\"> </span><span style=\"color: #81A1C1\">=</span><span style=\"color: #D8DEE9FF\"> </span><span style=\"color: #B48EAD\">2</span><span style=\"color: #81A1C1\">;</span>
+<span class=\"hl\"><span style=\"color: #8FBCBB\">console</span><span style=\"color: #ECEFF4\">.</span><span style=\"color: #88C0D0\">log</span><span style=\"color: #D8DEE9FF\">(</span><span style=\"color: #D8DEE9\">a</span><span style=\"color: #81A1C1\">+</span><span style=\"color: #D8DEE9\">b</span><span style=\"color: #D8DEE9FF\">)</span><span style=\"color: #81A1C1\">;</span></span>
+<span style=\"color: #8FBCBB\">console</span><span style=\"color: #ECEFF4\">.</span><span style=\"color: #88C0D0\">log</span><span style=\"color: #D8DEE9FF\">(</span><span style=\"color: #ECEFF4\">&apos;</span><span style=\"color: #A3BE8C\">shiki</span><span style=\"color: #ECEFF4\">&apos;</span><span style=\"color: #D8DEE9FF\">)</span><span style=\"color: #81A1C1\">;</span></code></pre>`
+  )
+})
+
+test('Line ranges are highlighted', async () => {
+  const highlighter = await getHighlighter({
+    theme: 'nord'
+  })
+  const code = `let a = 1;
+let b = 2;
+console.log(a+b);
+console.log('shiki');
+`
+  const out = highlighter.codeToHtml(code, 'js', { highlightLines: ['1-3'] })
+  expect(out).toBe(
+    `<pre class="shiki highlighted" style="background-color: #2e3440"><code><span class="hl"><span style="color: #81A1C1">let</span><span style="color: #D8DEE9FF"> </span><span style="color: #D8DEE9">a</span><span style="color: #D8DEE9FF"> </span><span style="color: #81A1C1">=</span><span style="color: #D8DEE9FF"> </span><span style="color: #B48EAD">1</span><span style="color: #81A1C1">;</span></span>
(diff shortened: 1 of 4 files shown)


aulneau:
I have my own fork that supports the ability to parse comments, abd if a line contains a comment that says singularly `highlight`, it will add a class to the line and remove the comment

packages/shiki/src/__tests__/simple.test.ts
```diff
@@ -7,3 +7,60 @@ test('Nord highlighter highlights simple JavaScript', async () => {
...
+  })
+  const code = `
+let a = 1;
+let b = 2;
+console.log(a+b);
+console.log('shiki');
+`
+
+  const out = highlighter.codeToHtml(code, 'js', { highlightLines: [2, 4] })
+  expect(out).toBe(
```
octref, line 23: You can use `toMatchSnapshot`.

packages/shiki/src/highlighter.ts
```diff
@@ -14,6 +14,16 @@ export interface HighlighterOptions {
...
 }
 
+export interface HtmlOptions {
+  // Pass an array of lines and line ranges (strikes like "4-18")
+  highlightLines?: (string | number)[]
+  addLines?: (string | number)[]
+  deleteLines?: (string | number)[]
+  focusLines?: (string | number)[]
+  // When debugColors is true, include token scope info in the HTML as data attributes
+  debugColors?: boolean
```
octref, line 24: Wondering if this is useful at all. Maybe only add it if someone requests?

packages/shiki/src/highlighter.ts
```diff
@@ -14,6 +14,16 @@ export interface HighlighterOptions {
   langs?: ILanguageRegistration[]
 }
 
+export interface HtmlOptions {
+  // Pass an array of lines and line ranges (strikes like "4-18")
+  highlightLines?: (string | number)[]
+  addLines?: (string | number)[]
+  deleteLines?: (string | number)[]
```
octref, line 21: What's the usecase for `deletedLines`?

octref (changes requested):
Thanks, good work 👍 

Although I'd prefer something like this as the option:

```ts
interface LineOption {
  line: number
  classes?: string[]
  skipTokenization?: boolean // default to false
}

interface HTMLRendererOption {
  lineOptions: LineOption[]
}
```

because:
- I don't think adding string number ranges to the API is useful. That should be handled by markdown plugin libraries.
- I find the class names a bit arbitrary. I'd like to let users pick their own classnames.

packages/shiki/src/highlighter.ts
orta: I wonder if it's for something like:

```diff
+ before
- after
```

octref: But wouldn't writing that inline be weird? I'd imagine someone highlight with the `diff` grammar.

packages/shiki/src/highlighter.ts
dceddia: Ah yeah I probably should've split this out as a separate feature/PR. I can do that.

This was very handy to have when I was customizing a color theme though! Before this, I was pretty much grepping through the color theme JSON for hex values and hoping that I was changing the right instance of that color. Having the scope names on the HTML elements made it much easier to track down the right colors to change.

Maybe there are better ways to do that, using some other tool to customize the color scheme. But I wanted to see them in context on the blog with as fast of a feedback loop as I could manage.

packages/shiki/src/highlighter.ts
dceddia: Yeah this was pretty much because I wanted a JSX-highlighted diff - check out [this example](https://daveceddia.com/react-confirmation-modal-state-machine/#display-the-error). I wasn't sure if Shiki supported diff combined with other grammars (jsxdiff, diffjsx, jsx+diff, something like that)... does it? I think I tried a few things but couldn't get it working.

From the related issue #74 here's an example of [what my Markdown looks like](https://github.com/shikijs/shiki/issues/74#issuecomment-681168224) for this. I've got comments at the top specifying the lines, and I strip those off at build time so that I can pass line numbers to Shiki. I toyed with the idea of using the diff markers +/- as line prefixes but it was just too painful to write out code that way 😄

dceddia:
I like your `LineOption` idea a lot! It's much cleaner and more general-purpose 👍 And the `skipTokenization` property will take care of the `deleteLines` case. The name `LineOption` doesn't really give a clue to what it's useful for but I can't really think of a better name.

The add/highlight/delete/focus attributes felt a bit hacky to me too. It started with just the highlightLines, and as it grew, it felt a bit "off" but I couldn't put my finger on it.

I think the `LineOption` would also extend nicely for another thing I wanted to add, the ability to highlight a range of characters within a line (say, for underlining a word or a changed part).

packages/shiki/src/renderer.ts
```diff
@@ -1,31 +1,76 @@
...
   }
   html += `<code>`
 
-  lines.forEach((l: IThemedToken[]) => {
-    if (l.length === 0) {
-      html += `\n`
-    } else {
+  lines.forEach((l: IThemedToken[], index: number) => {
+    const lineNo = index + 1
+    let highlightClasses = ''
```
codepunkt, line 45: I have another PR open (#77). Original idea behind the PR was to wrap every line in a `span` so i could easily add line numbers via CSS counters.

Due to the comments of @octref on that PR, i suggest calling this variable `lineClasses` instead of `highlightClasses` and start with a default class of `line`.

By doing so, we'd always wrap lines in a span, and that span would at least have the "line" class - which would obsolete my PR.

```javascript
let lineClasses = [
  'line',
  highlightedLines.has(lineNo) && 'hl',
  addLines.has(lineNo) && 'add',
  deleteLines.has(lineNo) && 'del',
  focusLines.has(lineNo) && 'foc',
].filter(Boolean).join(' ')

html += `<span class="${lineClasses}">`

if (l.length > 0) {
  l.forEach(token => {
    html += `<span style="color: ${token.color || options.fg}">${escapeHtml(
      token.content
    )}</span>`
  })
}

// Newline goes before the close, so that display:block on the line will work
html += `\n</span>`
```

packages/shiki/src/highlighter.ts
octref: I think it's better to just open an issue first, describe your use case, and we can see if we have a better way to do this or we have to add an option.

packages/shiki/src/highlighter.ts
octref: Can we start a separate issue for that feature, and not include it here for now?

I think what you have done is nice, but since highlighting is fast and static asset generation doesn't get rerun a lot, I think it's ok to highlight 2 duplicate files so the deleted lines can also have some color.

For example:

```
console.log(`
+  bar
-  foo
`)
```

We could highlight
```js
console.log(`
  foo
`)
```

and

```js
console.log(`
  bar
`)
```

We can just drop `skipTokenization` for now.

packages/shiki/src/renderer.ts
octref: > lineClasses instead of highlightClasses and start with a default class of line

👍  for that, but maybe as `LineOption.classes`, see top review comment.

sv3nk:
Hey everyone,

really excited for the idea of highlighting lines with shiki.
Is this currently being worked on or was something like it already included?

m1guelpf:
Are there any updates on this?

aarondfrancis:
Also really looking forward to this!

octref:
Please rebase and reopen if you still have time to dig into this. Thanks 🙏

mariusschulz:
@octref, I've submitted PR #259 and implemented a simplified version of this feature that supports highlighting individual lines as well as line ranges. Would love some feedback :)

PR #87, opened 2020-09-02
update readme: add blockstack

aulneau:
We are using shiki at blockstack for our docs: https://docs.blockstack.org, in combo with mdx, next.js, remark/rehype :)


octref:
Sure, thank you!

octref (approved):

PR #92, opened 2020-09-17
fix(readme): add destructure import example

jefrydco:
Since in `index.ts` we don't export any default module,

https://github.com/shikijs/shiki/blob/58cb92cd9427a97d5b4f89e3cec8e12290fcb793/packages/shiki/src/index.ts#L1-L6

it would raise an error like this, if we use `import shiki from 'shiki'` syntax.

```
The requested module 'file:///<some-path>/node_modules/shiki/dist/index.js' does not provide an export named 'default'
```

I think it's better to add destructure way to import `shiki`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ npm i shiki
 
 ```js
 const shiki = require('shiki')
+// import { getHighlighter } from 'shiki'
 
 shiki
   .getHighlighter({


octref:
Hmm, maybe I should export * as default, what do you think?

octref:
Sorry but I don't really think it clarify anything. TS users should be able to figure this out themselves so I'm removing it.

PR #93, opened 2020-09-17
Fix #63 for vuepress plugin

### LINKED ISSUE #63: Wrong HTML escape sequences in demo

yzhang-gh:
Copied from

https://github.com/shikijs/shiki/blob/58cb92cd9427a97d5b4f89e3cec8e12290fcb793/packages/shiki/src/renderer.ts#L39-L49

### DIFF
--- a/packages/vuepress-plugin/index.js
+++ b/packages/vuepress-plugin/index.js
@@ -14,10 +14,22 @@ module.exports = (options, ctx) => {
     chainMarkdown(config) {
       config.options.highlight((code, lang) => {
         if (!lang) {
-          return `<pre><code>${code}</code></pre>`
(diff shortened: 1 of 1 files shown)


octref (approved):
Ahh I missed escaping here. Thanks 👍

PR #96, opened 2020-09-27
add note on where to preview themes

sw-yx:
just adding a note since its a natural question to ask :)

### DIFF
--- a/packages/themes/README.md
+++ b/packages/themes/README.md
@@ -33,6 +33,8 @@ export type Theme =
   | 'solarized-light'
 ```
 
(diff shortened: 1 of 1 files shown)


octref:
Is that technically correct though? I cannot preview dark-plus or light-plus, for example.

sw-yx:
ah. oops. perhaps we can say you can preview *some* of these themes?

also wondering what are your thoughts for allowing people to bring their own themes?

octref:
@sw-yx That is already possible, see https://shiki.matsu.io

image

octref (approved):
LGTM, thanks 👍

Issue #91, opened 2020-09-16
RFC: Browser support & API changes


antfu:
Currently, `shiki` can only work on Node.js. There is the common usage for using a syntax highlighter right in the browser for dynamic contents or building a playground. This RFC is trying to make some changes to the current implementation to make shiki able to run on both Node and browsers isomorphically. https://github.com/shikijs/shiki/issues/22

I have made a hack build to replace `fs.readFileSync` with a sync `XMLHttpRequest` to fetch the data over http. You can try it at https://antfu.github.io/shiki-browser/. The main problem is that shiki will load all the grammars and themes in `getHighligter` which will cause high & redundant traffic & load cost in the browser.

### Internal Changes

Wrap the current `fs.readFileSync` with an isomorphic loader like

```ts
const isBrowser = typeof window !== 'undefined'

export async function loadFile(path) {
  if (isBrowser) {
	// load from related path or cdn
  } else {
    // load from fs
  }
}
```

### API Changes

New field `themes` and `languages` for `getHighlighter` to explicitly define what should be loaded. If not provided, they will work the same as we currently have.

```ts
await shiki.getHighlighter({
  // default theme
  theme: 'nord',

  // themes to load, optional
  themes: ['nord', 'github-light'],
  // languages to load, optional in Node, required in Browser
  languages: ['js', 'ts', 'html', 'vue'],
})
```

For users using bundlers like Webpack to build their app, they could make the loading by the bundler instead of CDN.

```ts
import shiki from 'shiki'
import nord from 'shiki-themes/data/nord.json'
import jsGrammar from 'shiki-languages/data/js.tmGrammear.json'
import htmlGrammar from 'shiki-languages/data/html.tmGrammear.json'
import customGrammar from 'path/to/custom.tmGrammer.json'

await shiki.getHighlighter({
  theme: nord,

  themes: [nordTheme],
  languages: [jsGrammar, htmlGrammar, customGrammar]
})
```

Also, add the ability to load grammars/themes incrementally for the highlighter

```ts
await highlighter.loadTheme('nord') // or a tm theme json
await highlighter.loadLanguage('wenyan') // or a tm grammer json
```

A third argument (optional) for `codeToHtml` to explicitly change the theme. Should raise an error if the provided theme is not loaded. This would be also useful for https://github.com/shikijs/shiki/issues/33

```ts
highlighter.codeToHtml(`console.log('shiki');`, 'js', 'min-light') 
```

### Build

Providing `umd` and `esm` builds.

[...]

PR #109, opened 2020-12-14
Refactor: Isomorphic

antfu:
Following this RFC: https://github.com/shikijs/shiki/issues/91

DEMO: https://codepen.io/antfu/pen/zYKNeWN

Try out now: `npm i @antfu/shiki` or `https://unpkg.com/@antfu/shiki`

## Changes
- Packages `shiki-languages` and `shiki-themes` get deprecated, languages and themes is not shipped with the `shiki` package
- Use rollup to build esm,cjs,iife bundle
- `getHighlighter()` internal changed from class to functional
	- New options `themes` allowing to load multiple themes asynchronously on initialization and being able to use them synchronously later.
	- `renderToHTML()` and `codeToThemedTokens()` support a third argument for specify the theme
	- New `async loadLanguage()` and `async loadTheme()` to load the resources incrementally
	- New `getBackgroundColor()` and `getForegroundColor()` API.
- CDN Support (iife build)
	- Exported to global variable `shiki`
	- `shiki-renderer-svg` will extend the `shiki` variable and available at `async shiki.getSVGRenderer()`
	- Builds for UNPKG and jsDelivr work out-of-box, and can be used for any CDN with `setCDN()`
- Use `lerna publish` to do the version bumping and release (all packages will be the same version)
- Removed `plist` support as they are already normalized to json

## BREAKING CHANGE
- `codeToThemedTokens()` options moved from the third arg to fourth, leaving the third for specifying `theme`
- when `langs` option in `getHighlighter()` is specified, it will no longer include all the bundled languages, use `langs: [...BUNDLED_LANGUAGES, myCustomLanguage]` to have the previous behavior
- `loadTheme` is now async


orta (commented):
I've given this a clone locally and run through a lot of the changes. 

I think this is really well thought PR, and we should merge it. I don't know how free @octref is, so I will give this at least till the end of January for his feedback. If we don't get any, I'll merge this, move the TypeScript website to it and if that all goes well then I'll ship the npm module.

After this PR, shiki should ideally be at v0.9.x - with shiki in the browser as doable then I don't know what else should be blocking hitting 1.0 and getting it well documented for others to use.

orta (approved):

43081j:
Can you add `exports` to your package manifests? Things are moving away from `module` and what not. Webpack also supports this, and possibly rollup, so it could be good to have it specified.

Like here:
https://nodejs.org/api/packages.html#packages_conditional_exports
https://webpack.js.org/guides/package-exports/

Also, what if I don't want to use a CDN? what is the usage here? is it expected we would load the file ourselves and pass the resulting object in for a theme for example?

Could you possibly rather than rely solely on CDNs, allow us to pass a path in to a theme/lang which the loader can then deal with?

packages/shiki/src/registry.ts
```diff
@@ -0,0 +1,64 @@
...
+      return this._resolvedThemes[theme]
+    } else {
+      return theme
+    }
+  }
+
+  public async loadTheme(theme: Theme | IShikiTheme | string) {
+    if (typeof theme === 'string') {
+      if (!this._resolvedThemes[theme]) {
+        this._resolvedThemes[theme] = await fetchTheme(`themes/${theme}.json`)
```
43081j, line 26: presumably this means if we want to load a theme by name, it must always exist in `themes/` which often won't be true. can it be changed to allow us to specify the exact path ourselves?

antfu: You can call the `loadTheme` API first and then passed the resolved theme object to it. You can also use your own loader to load your grammar/theme json and pass them to shiki, this is how you use it without CDN.

43081j: do you not think it would be a quick win to allow that to be deal with by the loader too? you already have the logic for CDNs and it would be identical logic almost for your own paths. it'd save consumers having to write pretty much the same logic in their codebases. just a thought

antfu: Not sure if I got your points, can you show me some examples? Thanks.

43081j: this is making a strong assumption that themes and languages always exist at `themes/` and whatever directory you ship languages in.

but this isn't true for many consumers due to bundling. it is increasingly likely that projects will split bundles and use dynamic imports these days, which means its possible the themes may well exist in production as individual files but the bundle which loads them may no longer be relatively in the same place it was in this repo.

so my suggestion would be one of the following:

```ts
// ideal but difficult for you because you wouldn't know which string is a path, and which is a theme name except by pattern
loadTheme('./my-own-theme.json');

// alternatively, make the assumed path configurable
registry.themesRoot = '/foo/bar';
loadTheme('test'); // results in '/foo/bar/test.json'
```

of course where `themesRoot` exists in config, so wherever you initialise shiki.

43081j: for example, one of our main repos uses webpack but _does not_ use non-standard imports, i.e. it sticks to native supported imports. which means no magical webpack imports (like json imports, etc).

in our case, we would ship the languages we want in an assets directory somewhere and load them up from the bundle which will be elsewhere.

in that case we'd have to deal with the loading manually to avoid your hard coded `themes/` path. so it would be nice to just tell shiki where to find the themes, or just give a path to the theme we want.

packages/shiki/src/loader.ts
```diff
@@ -0,0 +1,202 @@
+import * as Onigasm from 'onigasm'
+import type { IOnigLib, IRawGrammar, IRawTheme } from 'vscode-textmate'
+import type { IShikiTheme } from './types'
+import { dirname, join } from 'path'
```
43081j, line 4: out of interest, how does this work in browsers? is rollup or something stripping it?

43081j:
One last note is that your ESM output will not be consumable by browsers, which is a shame seeing as that is the way forward, not IIFEs.

If you could output an actual ESM bundle, that would be so nice. It wouldn't be a big change to your rollup file, it just means bundling it _with all externals_. So we can do this:

```html
<script type="module">
import {getHighlighter} from 'shiki'; // assuming we setup an import map already
// do stuff
</script>
```

this is also where `exports` would come in useful (in package.json) and means we can consume shiki in web components (which is why im here in the first place :joy: )

packages/shiki/src/registry.ts
antfu: Yeah, that makes sense. We can have a logic to detect if it's a built-in theme or relative/absolute path. While I don't have much bandwidth on this atm, would you mind opening a PR against my fork so the changes can be reflected here? Thanks.

antfu:
Agreed that we can have a browser ESM build

packages/shiki/src/registry.ts
43081j: No worries, I'll have a look through it soon and see what I can come up with

orta:
@octref one last ping, I'll give this one last look over next week (deal with conflicts etc) and merge otherwise

43081j:
FYI i did open a PR against this PR with the suggested path option at antfu/shiki#1

that and a browser ES module would be pretty useful to have, if possible

antfu:
Oh sorry, @43081j I didn't see it. Somehow GitHub doesn't watch for forks automatically. I will have a look later and get back to you.

Will solve the conflicts as well. (the main branch has a GitHub action that fetches the file updates which cause the conflicts, shouldn't be a big problem tho)

43081j:
no worries, if you have a better approach in mind just let me know and ill rework it if needed 👍

orta:
Alright, tests pass and things seem to build fine - I've given the diff another read over and I think this is good to go.

43081j:
@antfu did this end up containing a browser esm build? or will it have to be done in another PR?

PR #112, opened 2020-12-26
Handle errors in pullGrammarsFromGitHub.js

mariusschulz:
If we fail to download a certain grammar, we still want to continue with the remaining ones.


orta:
Agree, thanks

PR #108, opened 2020-12-13
FIX: Wrong language config files (React & TypeScript)

VladimirMikulic:
Beginning with shiki version 0.2.0, React (JSX)
and Typescript (ts) highlighting broke.

The reason for this was introduction of basically **empty**
jsx.tmLanguage.json & ts.tmLanguage.json which shiki would
use instead of **already valid** javascript.tmLanguage.json
and ts.tmLanguage.json files if jsx & ts modes were specified, respectively.

This behaviour caused an issue in highlighting as seen in #105 & #107.

This change removes empty jsx.tmLanguage.json & ts.tmLanguage.json
and **aliases** jsx and ts modes to javascript.tmLanguage.json
and typescript.tmLanguage.json, respectively.

Resolves #105 & #107

PR #114, opened 2021-01-25
Mark Highligher.codeToHtml as non-nullable

### LINKED ISSUE #101: Mark Highligher.codeToHtml as non-nullable
Highlither.codeToHtml is marked as a nullable property, however, as far as I can tell in practice, it will never be null. 

https://github.com/shikijs/shiki/blob/master/packages/shiki/src/highlighter.ts#L91

Am I missing something, or can this be marked as non-nullable?

leafac:
Closes #101

/cc @octref @orta

### DIFF
--- a/packages/shiki/src/highlighter.ts
+++ b/packages/shiki/src/highlighter.ts
@@ -124,7 +124,7 @@ export interface Highlighter {
     lang: StringLiteralUnion<Lang>,
     options?: ThemedTokenizerOptions
   ): IThemedToken[][]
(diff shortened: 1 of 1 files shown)


orta:
Yeah, I also can't see a reason why this would be nullable. Let's do it.

leafac:
How about we release to npm a version with this and https://github.com/shikijs/shiki/pull/108 ?

leafac:
For future reference: This has been shipped as part of 0.9.0.

PR #130, opened 2021-02-15
fix: use correct grammar file for cpp

### LINKED ISSUE #127: Poor highlighting result for cpp code
Hi, I'm using shiki to highlight code in my website. But I've found that it behaves strangely when highlighting cpp code. Here are the highlighting results(using theme light+):

image

As you can see, the macro `#include`, and the keyword `for`, `int`, `if` are not properly colored. And the dark blue color of `array` is also pretty suspicious. The comment is not colored, either. Below is the result from vscode(screenshot from my editor) for contrast:

image

I've also tried other theme like dark+ and nord, but the problem still didn't disappear.

### Reproduction guide

Install the latest shiki, and use following code:

```js
const fs = require('fs')
const shiki = require('shiki')

shiki
  .getHighlighter({
    theme: 'light-plus'
  })
  .then((highlighter) => {
    const html = highlighter.codeToHtml(
      fs.readFileSync('test.cpp', 'utf-8'),
      'cpp'
    )

    const out = `
<title>Shiki</title>
${html}`

    fs.writeFileSync('index.html', out.trim())

    console.log('done highlighting index.js')
  })

```

Example cpp code:

```cpp
#include <iostream>
using namespace std;

int main() {
  int array[100] = {0};
  for (int i = 0; i < 100; ++i) {
    if (array[i] == 0) {
      // do some thing...
    }
  }
}

```

Enter-tainer:
fix #127


orta:
Thanks for the notes in https://github.com/shikijs/shiki/issues/127 - your reasoning makes sense to me and the PR makes sense 👍🏻

PR #134, opened 2021-02-16
fix: wrap empty lines with span

### LINKED ISSUE #132: Always wrap lines in a span
I'm using shiki in my website, and i'm trying to add a line number to my code block. It depends on the span wrapped around the line. Something like:

```css
pre.shiki {
  counter-reset: shiki-line-number;
  /* whenever found a <pre class="shiki">, reset counter to zero */
}

code span.line:before {
  counter-increment: shiki-line-number;
  content: counter(shiki-line-number);
  /* everytime we found a new line in code block, increment the counter, and set the line number to its current value */
}
```
However, I've found that empty lines has no spans wrapped around them. And the result it pretty strange(empty lines have no line numbers):

image

While I'm trying to add code to wrap empty lines, I found that "do not wrap empty lines" seems to be intended.
(See [`078b489` (#77)](https://github.com/shikijs/shiki/pull/77/commits/078b4894a54c0af3f4221ae7f1308933018a2869#diff-919ab74d473366fabe3f5efd32806e5e8af7202cc565a46a71bb57bb303eb33eR21).)

So, my question is, why we need to skip empty lines? If this is not intended, I'll send a pr to fix it.

Enter-tainer:
close #132

### DIFF
--- a/packages/shiki/src/renderer.ts
+++ b/packages/shiki/src/renderer.ts
@@ -18,17 +18,13 @@ export function renderToHtml(lines: IThemedToken[][], options: HtmlRendererOptio
   html += `<code>`
 
   lines.forEach((l: IThemedToken[]) => {
-    if (l.length === 0) {
-      html += `\n`
-    } else {
-      html += `<span class="line">`
-      l.forEach(token => {
-        html += `<span style="color: ${token.color || options.fg}">${escapeHtml(
-          token.content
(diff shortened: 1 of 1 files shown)


orta:
Independently of this PR, I did the same thing for the twoslash shiki renderers today. So, I think this is probably the right way to go 👍🏻

PR #147, opened 2021-04-06
feat: add sparql and turtle grammars

stefanprobst:
this PR adds SPARQL and Turtle RDF grammars (incl. samples).

the formatting changes in `packages/shiki/src/languages.ts` come from the prettier pre-commit git hook afaict - happy to revert those if you want.


octref (approved):
LGTM 👍  I fixed three things on master:
- Auto-format all incoming JSONs (so there won't be format changes)
- Let prettier ignore languages.ts (I want alias to be on the same line as original id)
- Update the readme — `yarn update:grammars` should come after adding the sample, so it adds `samplePath`. I did that for your branch.

PR #149, opened 2021-04-18
Update themesPath and languagesPath earlier

banga:
The calls to `loadThemes` and `loadLanguages` used to occur before these options were applied, which meant that if you didn't have themes/languages in their default paths, `getHighlighter` would always fail.


octref:
Good catch, thank you 🙌

PR #151, opened 2021-04-23
Don't call `setTheme` if `_theme` hasn't changed

banga:
`setTheme` can be an expensive operation because it parses the raw theme
every time it's called:
https://github.com/microsoft/vscode-textmate/blob/9e3c5941668cbfcee5095eaec0e58090fda8f316/src/main.ts#L93-L95

In https://github.com/banga/git-split-diffs/, I invoke
`codeToThemedTokens` many times, one line at a time, so the cost of
`setTheme` adds up. In my local profiling, this was taking about 25% of
the time spent in `codeToThemedTokens`. I realize this is probably not
the intended way to use this function, but there isn't an easy
alternative for syntax highlighting individual lines in a diff.

Before
image

After
image

### DIFF
--- a/packages/shiki/src/highlighter.ts
+++ b/packages/shiki/src/highlighter.ts
@@ -53,6 +53,7 @@ export async function getHighlighter(options: HighlighterOptions): Promise<Highl
 
   const themes = await _registry.loadThemes(_themes)
   const _defaultTheme = themes[0]
+  let _currentTheme: IShikiTheme | undefined
   await _registry.loadLanguages(_languages)
 
   function getTheme(theme: IThemeRegistration) {
@@ -60,7 +61,10 @@ export async function getHighlighter(options: HighlighterOptions): Promise<Highl
     if (!_theme) {
       throw Error(`No theme registration for ${theme}`)
     }
-    _registry.setTheme(_theme)
+    if (_currentTheme !== _theme) {
+      _registry.setTheme(_theme)
+      _currentTheme = _theme
+    }
     const _colorMap = _registry.getColorMap()
     return { _theme, _colorMap }
   }


banga:
cc @octref in case you haven’t seen this PR. Thanks!

octref (approved):
Thanks for the contribution! The only thing I'm not 100% happy about is we are doing object comparsion, not name comparison, so the code here is a bit dirty. I'll open another issue to make `name` required in `IShikiTheme` and change the comparison here.

banga:
Makes sense, thank you!

octref:
Hey @banga, I just saw your repo on HN frontpage, congrats!
I'm wondering if I can mention it in "related projects" section. Additionally, do you have any thoughts on making a terminal renderer for shiki? I have had this thought but haven't given it a try. Would love to know if you have run into any bumps there.

banga:
Thanks and thank you for making this project!

> I'm wondering if I can mention it in "related projects" section. 

Yes of course!

> Additionally, do you have any thoughts on making a terminal renderer for shiki? I have had this thought but haven't given it a try. Would love to know if you have run into any bumps there.

Yeah that sounds useful. The initial implementation was actually really easy. I used [chalk](https://github.com/chalk/chalk) which already supports hex colors and converts them to appropriate ANSI sequences for the current terminal.

I eventually had to do more work for supporting alpha in colors (so I could e.g. highlight inline changes without overwriting the underlying syntax highlighting), so I implemented https://github.com/banga/git-split-diffs/blob/main/src/SpannedString.ts. It felt like I was sort of reinventing html, but I needed something tailored to my use-case of highlighting spans in a line, one line at a time, and then flattening it into styled segments.

Slightly unrelated, but one thing I'm invested in for my project is speeding up startup time – we currently spend a bunch of time in parsing the grammars and themes to in-memory structures. I know very little about the internals of vscode-textmate, but it seems that there might be a way to cache this parsed version on disk and quickly load that in instead? That would be helpful in enabling more CLI apps to use this project, since developers tend to be really sensitive to perf issues in those.

octref:
> cache this parsed version on disk and quickly load that in instead

@banga I just came across this issue again. I currently don't have time to look into this yet. Would be great if that could happen.

banga:
> > cache this parsed version on disk and quickly load that in instead
> 
> @banga I just came across this issue again. I currently don't have time to look into this yet. Would be great if that could happen.

I tried this today, but it was actually slightly (~6%) slower in my tests. This is what I tried: https://github.com/microsoft/vscode-textmate/compare/main...banga:main.

Separately, I noticed that after updating to `v0.9.5`, the startup time got much slower. This seems to be because we now default to loading all `BUNDLED_LANGUAGES` if you don't specify a non-empty `langs` option. Perhaps some people are running into this?

octref:
> Separately, I noticed that after updating to v0.9.5, the startup time got much slower. This seems to be because we now default to loading all BUNDLED_LANGUAGES if you don't specify a non-empty langs option. Perhaps some people are running into this?

I just opened https://github.com/shikijs/shiki/issues/200.

> I tried this today, but it was actually slightly (~6%) slower in my tests. This is what I tried: microsoft/vscode-textmate@main...banga:main.

To avoid losing contexts, do you mind opening a new issue so we can track this? Plan to fix this, but needs more digging into it.

PR #162, opened 2021-05-17
Allow dropping the theme option in the highlighter's getTheme interface

orta:
`getTheme` supports passing an `undefined` to grab the default but the types doesn't cover that case, this makes the param optional to account for that.


orta:
Looks like the master branch is red ( I wondered if it was just my build)

octref:
LGTM, sorry for the slow response.

PR #140, opened 2021-03-18
Some changes from my fork

Vap0r1ze:
I made a fork that has a few fixes for things that were causing issues with a project of mine. Figured I'd open a PR to see which changes you'd be willing to merge:
  - corrected the order in which `settings` were `concat`ed in themes that `include`d others
  - import [json5](https://github.com/json5/json5) to parse json files with comments
  - preserved the `colors` property in themes
  - exposed [Highlighter#getTheme](https://github.com/shikijs/shiki/blob/c838ebcf7e7ab419fc10a6a9cfc83d9236c8e1e0/packages/shiki/src/highlighter.ts#L58-L66)

PR #164, opened 2021-05-19
add vpc-shiki to example projects

Vap0r1ze:
You suggested this be done in #140


orta:
Looks good to me, thanks

PR #174, opened 2021-07-17
Adds DreamMaker language support

4310V343k:
For the sample i just yanked the code from an actual game (https://github.com/tgstation/tgstation)


orta:
Dream Maker was my first language, I still have a copy of the original blue book around

PR #189, opened 2021-08-02
feat: support web workers

### LINKED ISSUE #188: Cannot use in Web Worker
Hi, when using `shiki` in a Web Worker, `window` is not defined, therefore this line https://github.com/shikijs/shiki/blob/master/packages/shiki/src/loader.ts#L7 is `false`.

And when `isBrowser` is `false`, it tries to require things at runtime and it just doesn't work.

Is there a way to detect in `shiki` if it's running in a Web Worker and set `isBrowser` to `true` ?
I can confirm everything runs smoothly in Web Workers when `isBrowser` is `true` .

If not, we may need to provide a function that manually set the `isBrowser` variable :)

kevinmarrec:
Resolves #188 

These changes aim to make `shiki` working in web workers :)

Side note : `WebWorker` is needed in `tsconfig.json` to make `WorkerGlobalScope` available on `self`.

### DIFF
--- a/packages/shiki/src/loader.ts
+++ b/packages/shiki/src/loader.ts
@@ -5,7 +5,7 @@ import type { IOnigLib, IRawGrammar, IRawTheme } from 'vscode-textmate'
 import type { IShikiTheme } from './types'
 
 export const isWebWorker =
-  typeof self !== 'undefined'&&
+  typeof self !== 'undefined' &&
   typeof self.WorkerGlobalScope !== 'undefined'
 
 export const isBrowser =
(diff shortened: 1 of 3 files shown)


antfu (approved):

orta (approved):
Agree, this looks good to me

orta:
CI fail is legit (it's typescript, just throw a `// @ts-expect-error` as it's a safe call


```
    packages/shiki/src/loader.ts:9:15 - error TS2339: Property 'WorkerGlobalScope' does not exist on type 'Window & typeof globalThis'.
typeof self.WorkerGlobalScope !== 'undefined'
```

kevinmarrec:
@orta Indeed, I didn't notice that Github Actions needed maintainer approval, so I only saw green tick next to my commit, tests only have been triggered afterwards and failed.

I pushed a commit where I added `WebWorker` to the `lib` array of the main `tsconfig.json` which may fix the tests. At least, it works on my computer locally with this change (`yarn build && yarn test`).

orta:
Confirming that this worked in a nextjs app with webpack 5

kevinmarrec:
Confirming that this also worked in a nuxtjs app with webpack 4

PR #192, opened 2021-08-04
chore: update deps

antfu:
PR for CI


orta:
I have access to netlify, but maybe the `netify.toml` allows setting the node version?

antfu:
Yes, I was interpreted by other stuff, should be good now

antfu:
Btw, netlify does not triggered since the change for branch name. Will need a re-link I think.

orta:
oh boo

orta:
That will require @octref - I think we should tread on and I'll make an issue about it

PR #198, opened 2021-08-13
Fix Vue template highlighting

yyx990803:
For some reason `vue-html` isn't registered in languages, which causes `<template>` blocks in Vue files not being properly highlighted.

Adding this also enables highlighting a piece of Vue template directly (without having to nest it inside `<template>` as in SFCs).


octref:
LGTM, thanks!

yyx990803:
Can we publish a new version of shiki with this change?

antfu:
@yyx990803 released as v0.9.7

PR #203, opened 2021-08-15
Remove "const" from publicly exported enums

### LINKED ISSUE #172: Bug: Shiki cannot be used with --isolatedModules flag
This was introduced between 0.9.3 and 0.9.5. TypeScript is complaining about `FontStyle` here.

```
node_modules/shiki/dist/index.d.ts:269:68 - error TS2748: Cannot access ambient const enums when the '--isolatedModules' flag is provided.   

269 export { languages as BUNDLED_LANGUAGES, themes as BUNDLED_THEMES, FontStyle, Highlighter, HighlighterOptions, HtmlRendererOptions, ILanguageRegistration, IShikiTheme, IThemeRegistration, IThemedToken, Lang, Theme, getHighlighter, fetchTheme as loadTheme, renderToHtml, setCDN, 
setOnigasmWASM };
```

Gerrit0:
This makes it possible to use Shiki with projects that set the "isolatedModules" flag for improved compiler performance.

Closes #172

### DIFF
--- a/packages/shiki/src/stackElementMetadata.ts
+++ b/packages/shiki/src/stackElementMetadata.ts
@@ -40,7 +40,7 @@ export const enum TemporaryStandardTokenType {
   MetaEmbedded = 8
 }
 
-export const enum FontStyle {
+export enum FontStyle {
   NotSet = -1,
   None = 0,
   Italic = 1,


1chiSensei:
Removing the `const` from it makes it not usable in runtime since removing it causes the enum to be a type.

Gerrit0:
Huh? That makes absolutely no sense. Enums in TypeScript have both types and values. Whether or not they result in runtime code is determined by two things.

1. Whether or not they are `const`
2. The [preserveConstEnums](https://www.typescriptlang.org/tsconfig#preserveConstEnums) setting

If an enum is *not* `const` it will *always* result in a runtime object being created. If it is `const`, a runtime object will be created only if `preserveConstEnums` is set.

octref (approved):
LGTM 👍

octref:
I'm positive both will be usable in runtime. I don't see any downside for changing it.

StoneCypher:
well, it's not type checkable anymore, and the .d.ts is no longer correct

your original code was close to correct, and would have been correct if you extracted the array from the type

the real problem was that your downstream consumer is attempting to describe himself as an `--isolatedModule`, but he's also importing an enum from you.  What `--isolatedModule` actually does is to prevent the importation of types, and importing an enum is importing a type.

He had a simple bug.  This fix isn't correct, @octref 

He had three options.

The correct option is to remove `--isolatedModule`, because it's not an isolated module.  That's a performance hack.  Unfortunately he seems to believe that it's a strictness thing, and refuses to turn it off on those grounds, despite that it is actually incorrect.

There are two other options.

One, he could cast the type to string, and make the comparison there.  

Another, he could specify the type himself, in his own .d.ts.

This should be a string alternation as const, like it originally was, and the array should be extracted from the type.  Until it's returned to `as const`, the array cannot be extracted from the type, and this type cannot be unified.

Issue #33, opened 2020-04-10
Question: Recommendations on how to handle dark mode


orta:
👋 I'm thinking about how to handle dark and light themes, https://github.com/microsoft/TypeScript-Website/issues/342 eventually I'm going to need to figure out how to make a dark mode which uses css classes because these poor dark mode folks have the same code color schema:

<img width="1458" alt="Screen Shot 2020-04-09 at 9 39 25 PM" src="https://user-images.githubusercontent.com/49038/78954618-b2bd4980-7aaa-11ea-8ff2-5501b5128e4f.png">

<img width="1458" alt="Screen Shot 2020-04-09 at 9 39 17 PM" src="https://user-images.githubusercontent.com/49038/78954646-c36dbf80-7aaa-11ea-9e30-e501a6eee029.png">

So, I'm wondering what you think might be the best way to do it (and whether it's something I should do upstream back here) 

I have two ideas:

 - Have a single theme, convert the colors hex to css classes
 - Try match two separate themes together

giovanicascaes:
Does your first idea mean using a single style that looks good both in light and dark mode?

Also, I'd like to help implementing this, is it something lower in complexity?

orta:
I think of it as being: 

- Shiki returns "#ffeeaa" for a token
- My app makes a sha of it: "#ffeeaa" -> "asd12" 
- My app sets the class of the token to be "asd12"
- We keep track of all colors which were assigned and spit out a single css file for all those colors
-  That CSS file can use `prefers-color-scheme: dark`

giovanicascaes:
It is not an elegant solution but code conteiners could be dark even in light theme. Check Dan Abramov's overreacted.io (https://overreacted.io/goodbye-clean-code/, for example). It looks good for me to read, both in light and dark theme. That seems the easiest possible solution. At least could bring some fresh to the eyes at night until the ideal solution could come up.

orta:
Yep - that's a reasonable workaround to not have this problem. That's totally doable for people, but I'm not planning on making that design compromise

giovanicascaes:
Yes, I understand. I'm not proposing defining as design choice, but it is a better solution than the current one. I believe devs feel more comfortable reading dark code during the day than light code at night.

Microflash:
[...]

PR #212, opened 2021-08-26
Add "css-variables" theme, support dark mode

FredKSchott:
- Adds local "css-variables.json" theme
- Adds documentation for this theme, and dark mode support in general
    - (feel free to make any changes to docs as you see fit!)
- Closes (potentially) #33 ?


FredKSchott:
@orta alright, I got a new implementation working for this PR, now assuming the VSCode bug does not get fixed. If we think it might get fixed, we can revert back to the original theme.

The new implementation is essentially a more controlled take on @pveyes's original approach to hex->variable mapping. The new theme uses valid hex color code as unique identifiers (`#000001`, `#000002`, `#000003`, etc) to get around the VSCode limitation. Once the theme has been loaded and we have the color map (`_colorMap = ['#000001', '#000002', ...`) we can replace those with the matching CSS variable that we care about.

Let me know if this is acceptable. I don't _love_ doing the hex->var mapping, but since we have direct access to `_colorMap` after the theme is parsed the "fix" step is actually pretty lightweight. 

If this method isn't acceptable, then I think we're blocked by the limitation linked to above, and I'll probably just go ahead with this solution in user-land (already implemented in Astro in https://github.com/snowpackjs/astro/pull/1208)

antfu:
That's an interesting workaround! Nice idea! I am good with it 👍

orta:
I agree, given that it's all happening internally, I don't think it's too hacky of an answer

FredKSchott:
Great, glad this makes sense. In that case, the PR is ready to merge on my end!

orta:
Should be shipped as 0.9.9

octref:
Thanks for your contribution @FredKSchott! Please take a look at https://github.com/shikijs/shiki/issues/33#issuecomment-921229816 though.

PR #215, opened 2021-09-01
[css-variables] fix theme fg & bg

### LINKED ISSUE #214: css-variables theme doesn't apply variable to <pre> wrapper style attr
The new `css-variables` theme is rad 👍🏻  
I'm on `shiki@0.9.9`

I don't think the correct style is being applied to the `<pre>` wrapper

I get
```html
<pre class="shiki" style="background-color: #000002">...
```

I would expect
```html
<pre class="shiki" style="background-color: var(--shiki-token-background)">...
```

My highlighter instance looks like
```js
    theme: 'css-variables',
    langs: [
      'bash',
      'javascript',
      'json',
      'powershell',
      'python',
      'ruby',
      'toml',
      'yaml',
      {
        // a custom grammar
      }
    ],
  })
```

The rest of the theme is working as expected.

FredKSchott:
Resolves #214 

While testing the latest changes in the Astro repo I uncovered a small bug in #212 where the `foreground` and `background` styles were not being applied. @tbeseda ran into the same in #214.

The fix is that foreground and background are managed separately, outside of the color map. This PR adds these two properties to the "fix" step. Tests also updated.

### DIFF
--- a/packages/shiki/src/__tests__/__snapshots__/cssVars.test.ts.snap
+++ b/packages/shiki/src/__tests__/__snapshots__/cssVars.test.ts.snap
@@ -1,7 +1,7 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`The theme with css-variables renders correctly 1`] = `
-"<pre class=\\"shiki\\" style=\\"background-color: #000002\\"><code><span class=\\"line\\"></span>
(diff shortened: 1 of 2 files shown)


antfu (approved):

orta:
👍🏻

PR #218, opened 2021-09-02
Add CodeChalk to `Seen` section

a20185:
Hi @octref . We've implement a util based on shiki. It converts shiki tokenizer result to chalk calls, which enables syntax highlighting for outputing instruction codes in terminals.

I'm wondering if it's possible to add the tool to `Seen` part, thanks~


orta:
Thanks!

Issue #169, opened 2021-06-25
Performance doubled between 0.9.3 and 0.9.5


orta:
Timing for the test suite for Shiki Twoslash doubled when we updated (and then reverted with 
https://github.com/shikijs/twoslash/pull/38/commits/e30e47e0d7bc9dcd2283dd317a9d9b1054942e60 )

I don't have time for a look why, but just raising incase there's something obvious

octref:
Was there a misconfiguration on your side? I run pnpm test with both 0.9.3 and 0.9.5, and they took roughly the same time (7s).

Test Suites: 8 passed, 8 total
Tests:       39 passed, 39 total
Snapshots:   7 passed, 7 total
Time:        6.013 s

Test Suites: 8 passed, 8 total
Tests:       39 passed, 39 total
Snapshots:   7 passed, 7 total
Time:        6.02 s
Ran all test suites.

riasvdv:
I'm having a similar issue. Running the following script with node:

```js
shiki.getHighlighter({
    theme: 'nord',
}).then((highlighter) => {
    process.stdout.write(highlighter.codeToHtml('console.log(\"Hello world\")', 'js'));
});
```
Takes ~880ms on 0.9.3 and ~3s on 0.9.5

If I call `getLoadedLanguages` on 0.9.5, all languages seem to be loaded, is this intended? I can't see which languages are loaded on 0.9.3 as the function isn't available there

orta:
Yeah, looking at this again, after updating a second time https://github.com/shikijs/twoslash/pull/104

Before:

```
PASS packages/shiki-twoslash/test/twoslash-renderer.test.ts (7.515 s)
PASS packages/remark-shiki-twoslash/test/fixtures.test.ts (15.383 s)
```

After:

```
PASS packages/shiki-twoslash/test/twoslash-renderer.test.ts (12.045 s)
FAIL packages/remark-shiki-twoslash/test/fixtures.test.ts (26.313 s)
```

Gerrit0:
Spent a bit of time looking at this, and I can claim fairly confidently that 0.9.8 is slower than 0.9.3, though I'm not seeing times double.

```js
const { performance } = require("perf_hooks")
const shiki = require("shiki")
const fs = require("fs")
const { join } = require("path")
const code = fs.readFileSync(join(__dirname, "node_modules/shiki/dist/index.js"), "utf-8")

async function main() {
    const start = performance.now();
[...]

PR #222, opened 2021-09-05
Switch from json5 to vscode-jsonc

Gerrit0:
Possibly a partial resolution for #169.

As far as I can tell, #11's premise that we should use JSON5 because that's what VSCode uses was wrong. VSCode does support JSON with comments, but does not appear to support the full range of JSON5. This is a really good thing for us, because it means we can switch from JSON5's [really slow parser](https://github.com/json5/json5/issues/234) to jsonc-parser.

In a sample run from TypeDoc, we spent 27% of the total runtime in JSON5's parser. With this change, only 7% of the time is spent parsing. This still seems somewhat high to me, given that TypeDoc runs the TS compiler, which ought to be the majority of the run time, but it's an improvement!


Gerrit0:
... so why did that fail on CI and not locally? I ran those tests!

Gerrit0:
CI is breaking for an unrelated reason - CI updates grammars before running, and https://github.com/microsoft/vscode/commit/bd2493499dab8aafe8d216dfa4de7dafd04b4c94 appears to have broken the grammar update.

orta:
I can't rebase this because it's not ticked for allowing maintainers to edit, overall though, I like it - it's a dependency already in our tree so switching to it seems reasonable to me 👍🏻  and the fail is definitely unrelated, so I'm pretty good with that signal.

PR #227, opened 2021-09-13
Handle token colors with multiple font styles

iliana:
`FontStyle` is bitwise and themes can set multiple font styles simultaneously (e.g. https://github.com/theacodes/witchhazel/blob/5dc744fb0fb7210be1ccacc5863a5768644a8380/witchhazelhypercolor.tmTheme#L139-L151).

Tested locally with a converted version of Witch Hazel Hypercolor (as linked above).


octref:
Thanks for your contribution! Rebased and had to update the commit message as we just adopted Conventional Commits: #230

I'll add a test as well.

PR #228, opened 2021-09-13
Add licensing information to packages

JonasKruckenberg:
Howdy!
I noticed the packages in this monorepo were missing License files. This causes problems with many automated tools as your License file will not be included in the distributed rpm bundle. Thought I'd shoot you a PR to fix this 😉


orta:
Seems reasonable enough, the license meta is in the package json - but I can respect that some tools don't take that into account 👍🏻

orta:
Fail is unrelated 👍🏻

Issue #206, opened 2021-08-23
Custom <pre> class name?


FredKSchott:
In Astro, we'd love to power our new built-in `<Code>` component with Shiki (to eventually replace our current built-in `<Prism>` component). 

However, we would like to keep shikli as an implementation detail, and not leak the `class="shiki"` attribute onto the `<pre>` element if possible. 

Is this something that you / the team is interested in adding as a configuration value to `codeToHtml()` to customize the `<pre>` class name that gets added? If so, I'm happy to provide the PR.

Issue #220, opened 2021-09-03
Custom <pre> / <code> attributes


naiyerasif:
Please consider a way to add custom attributes on `pre` / `code` elements. Some usecases:

- **Keyboard navigation of codeblocks**: To make the codeblocks accessible for keyboard users, `tabindex = "0"` should be set on `pre` element so that they can scroll them with keyboard. As a short-term goal, we can add `tabindex = "0"` by default. In long-term, we can still keep `tabindex="0"` as default but allow people to customize it. More details can be found here: https://accessibilityinsights.io/info-examples/web/scrollable-region-focusable/
- **Adding [data-*] attributes** for customizing the behavior of codeblocks, e.g., adding a "Copy to Clipboard" button for specific codeblocks, displaying codeblock title, language, etc.

fengzilong:
I'm willing to work on a PR for this feature

PR #229, opened 2021-09-16
Add `wrapperClasses` param to `codeToHtml`

tbeseda:
Related: #206 

### Motivation

Sometimes I'd like to add classes to Shiki's `<pre>` tag because my site uses utility classes (ala Tailwind). Additionally, it might be nice to opt-out of the `shiki` class in generated HTML.

### Implementation

I based this implementation on the suggestion in #206 and added an optional 4th param to `codeToHtml`: `wrapperClasses`.
Usage:
```js
const html = highlighter.codeToHtml(`console.log('test');`, 'js', 'nord', ['rounded-full', 'mx-auto'])
// <pre class="rounded-full mx-auto" style=...
```
And to remove all classes:
```js
const html = highlighter.codeToHtml(`console.log('test');`, 'js', 'nord', [])
// <pre style=...
```

### Drawbacks and Notes

1. The 3rd param for `codeToHtml` becomes required if the user wants to set classes with the 4th param 😐 
  i. passing an options object as the third param that could contain `theme` and `wrapperClasses` would be better, but that would be a breaking change to the API
1. If the user wants to keep the `shiki` class and add custom classes, they need to include "shiki"
  i. this seems fine to me and keeps the code simpler
1. I didn't include "extra win" from the original issue to add the theme name as a class
  i. this could be really helpful to target specific theme styles with CSS
  ii. Shiki could use a custom token in the array of classes: `['rounded-full', '$theme']`

Let me know what you think! Happy to make changes.

### DIFF
--- a/packages/shiki/src/highlighter.ts
+++ b/packages/shiki/src/highlighter.ts
@@ -123,14 +123,20 @@ export async function getHighlighter(options: HighlighterOptions): Promise<Highl
     return tokenizeWithTheme(_theme, _colorMap, code, _grammer, options)
   }
 
-  function codeToHtml(code: string, lang = 'text', theme?: StringLiteralUnion<Theme>) {
+  function codeToHtml(
+    code: string,
+    lang = 'text',
+    theme?: StringLiteralUnion<Theme>,
+    wrapperClasses?: string[]
+  ) {
     const tokens = codeToThemedTokens(code, lang, theme, {
       includeExplanation: false
     })
     const { _theme } = getTheme(theme)
     return renderToHtml(tokens, {
       fg: _theme.fg,
-      bg: _theme.bg
+      bg: _theme.bg,
+      wrapperClasses
     })
   }
 
--- a/packages/shiki/src/__tests__/__snapshots__/wrapper-classes.test.ts.snap
+++ b/packages/shiki/src/__tests__/__snapshots__/wrapper-classes.test.ts.snap
@@ -1,0 +1,5 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`empty wrapperClasses omits <pre> class attribute 1`] = `"<pre style=\\"background-color: #2e3440ff\\"><code><span class=\\"line\\"><span style=\\"color: #D8DEE9\\">console</span><span style=\\"color: #ECEFF4\\">.</span><span style=\\"color: #88C0D0\\">log</span><span style=\\"color: #D8DEE9FF\\">(</span><span style=\\"color: #ECEFF4\\">&#39;</span><span style=\\"color: #A3BE8C\\">test</span><span style=\\"color: #ECEFF4\\">&#39;</span><span style=\\"color: #D8DEE9FF\\">)</span><span style=\\"color: #81A1C1\\">;</span></span></code></pre>"`;
+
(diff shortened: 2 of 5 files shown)


packages/shiki/src/highlighter.ts
```diff
@@ -123,14 +123,20 @@ export async function getHighlighter(options: HighlighterOptions): Promise<Highl
     return tokenizeWithTheme(_theme, _colorMap, code, _grammer, options)
   }
 
-  function codeToHtml(code: string, lang = 'text', theme?: StringLiteralUnion<Theme>) {
+  function codeToHtml(
+    code: string,
+    lang = 'text',
+    theme?: StringLiteralUnion<Theme>,
+    wrapperClasses?: string[]
```
antfu, line 130: I guess at this point, we might want the third argument to be an options object for better extensibility and readability. WDYT?

tbeseda: Yes, once a method has > 1 optional arguments it's probably best to move to an options object. `codeToHtml` is pretty much the main interface for Shiki, so it would be a bigger breaking change, but it could help a lot in the long run

antfu: Could fallback to theme if it's a string for compatiblity.

tbeseda: So maybe overload the function with a union-type 3rd arg like `string | string[]`?
both of these would work:
```ts
codeToHtml(`console.log('test');`, 'js', 'nord', ['rounded-full', 'mx-auto'])
codeToHtml(`console.log('test');`, 'js', ['rounded-full', 'mx-auto'])
```
at least until Shiki decides it's time to make a breaking change and use an options arg.

FredKSchott:
+1 for the Astro use-case here. We use shiki internally and do not actually expose the `shiki` class name as an internal implementation detail. Currently, we fix this with a regex find-replace on the final HTML to remove it.

packages/shiki/src/highlighter.ts
octref: I agree with @antfu. We don't want too many non-essential options here.  It's pre 1.0 so we could make breaking changes. I just want to make those changes in one batch.

I also don't like too many compatibility union options. They confuse users. Better aim for a good API in 1.0 and then just make those breaking changes.

antfu: I guess currently we could have:

```ts
export interface CodeToHtmlOptions {
  theme?: StringLiteralUnion<Theme>,
  classes?: string
}

export function codeToHtml(
  code: string,
  lang = 'text',
  themeOrOptions: StringLiteralUnion<Theme> | CodeToHtmlOptions = {}
) {
  let theme
  let classes = 'shiki'
  if (typeof themeOrOptions === 'string') {
    theme = themeOrOptions
  }
  else {
  	theme = themeOrOptions.theme || theme
    classes = themeOrOptions.classes || classes
  }
}
```

octref: How about making `theme` the 3rd required option and `options` 4th optional param?

tbeseda: We don't pass the third param in our use case since we've used `loadTheme` before running `codeToHtml`. Of course, if my initial implementation landed as-is, we'd need to adjust anyway 🙃 

IMO a call like `codeToHtml(code, lang, { classes: ['rounded-full', 'mx-auto'] }` where I don't need to worry about which theme is going to be used is most elegant. but I also don't mind being more verbose.

I'm happy to help with any implementation preferred by maintainers, just wanted to get the ball rolling.
cheers

octref: @tbeseda You still need to specify a theme even after `loadTheme`, don't you?

tbeseda: @octref the theme is loaded and then passed to `getHighlighter`, which I believe makes it default when we use `highlighter.codeToHtml`

```js
// inside an async method that returns HTML
const theme = await shiki.loadTheme('./a-custom-theme.json')
  theme,
  langs: [...]
})

return highlighter.codeToHtml(code, lang)
```
Maybe this isn't intended? I haven't traced how the highlighter + registry works internally.

Gerrit0:
> Currently, we fix this with a regex find-replace on the final HTML to remove it.

At this point, why don't you implement your own renderer? It's a trivial thing to do with the themed tokens.

FredKSchott:
there's a lot of features in `renderer.ts` that I wouldn't want to fork/copy-paste and then risk getting out of sync with Shiki. Obviously if it ever comes to that that's not a bad option to have in our back pocket, but right now the regex-find-replace is oddly more stable.

orta:
Thanks for the PR! - :tada: this one is deprecated though because I merged #220 which would allow any sort of wrapper at a few other levels too 👍🏻

PR #243, opened 2021-10-09
feat: support custom elements for html renderer

### LINKED ISSUE #220: Custom <pre> / <code> attributes

fengzilong:
- [x] Add a test if possible

- [x] This PR fixes #220


kylebutts:
I'm a really big fan of this! This would make my workflow much easier than find and replacing things :-)

jrson83:
Please merge this PR!

a-trost:
This would be fantastic for being able to display what language a code block is showing.

fengzilong:
The merge conflicts were solved

orta:
OK, I'm making an executive call on this - I think this is a pretty solid abstraction, and allows for people to be able to work around many cases with a single set of tools. I think this is one we should have in WRT #220

PR #256, opened 2021-11-15
feat: path renderer

### LINKED ISSUE #255: Feature Request: shiki-renderer-svg serverless
`shiki-renderer-svg` relies on `playwright` to measure font size, which means that the server needs to install a whole browser(and many dependencies) to run that, making it's hard to run at a serverless service or alpine node.

tmkx:
- [x] Add a test if possible

- [x] This PR fixes #255 

I'm not sure it's suitable to make changes on `renderer-svg` directly(seems like a breaking change), so I made a copy of it.

The main changes are:
* use `opentype.js` to measure font width, so we need to pass a font file path, or font file buffer (~~could not use css font-family directly~~, fetch the font file and convert to ArrayBuffer first)
* embedded font, more portable, but not copiable 😢
* render to image using something like `node-canvas`, so we don't need to install a browser (not implemented yet)


octref:
Hey, thanks for your contribution! If possible I of course don't want to bundle a whole browser into the dependency.
I have tried both opentype.js and node-canvas before, however given the same fonts, they don't give me exactly the same width/height measurement as Chrome does, and the rendered SVG would be jumbled together. Have you tried to verify this?

tmkx:
@octref I've made a demo: https://shiki-path-renderer.herokuapp.com/

octref:
Sorry it took a while for me to get back to you. This is great work!

I did a comparsion of the SVG generated. The one above is `<path>`, the one below is `<tspan>`:

image

If I turn off subpixel antialiasing with `-webkit-font-smoothing: antialiased;`, they look almost identical:

image

>I'm not sure it's suitable to make changes on renderer-svg directly(seems like a breaking change), so I made a copy of it.

I'm down to scrape the current svg renderer and replace it with the OpenType implementation.
How about this – I can merge this PR as is and combine it with the `renderer-svg` one.
Is the main problem for you the API for importing font?

> render to image using something like node-canvas, so we don't need to install a browser (not implemented yet)

I'm looking into using resvg-js.

tmkx:
Hi @octref, I'm considering that this is a tradeoff:


| Renderer                 | Pros                        | Cons                                                         |
| ------------------------ | --------------------------- | ------------------------------------------------------------ |
| tspan (Headless Browser) | Copyable; SVG size is small | Heavy to run; Need font installed                            |
| path (OpenType)          | Easy to use; Portable       | CJK/Symbols support (Shows a blank if the character is missing in the selected font) |

* _Copyable_ is not so important when it shows as a `<image />`
* SVG path can be compressed using [\<use /\>](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/use)
* Unknown characters can be fallback to tspan

What do you think?

tmkx:
> I can merge this PR as is and combine it with the renderer-svg one.

That's great

octref:
@tmkx Looks great! But I think you can also use `opentype.js` to generate `tspan` as well. The headless browser was only included to measure font width as I didn't get a reliable measurement with server-side solutions, for example: https://github.com/samizdatco/skia-canvas/issues/4

octref:
Rebased on the latest main so the CI is working. I'll take a look later today.

tmkx:
rebased and tests passed on my computer.

octref:
Closing and reopening to see if the CI works

octref:
I'll fix the netlify CI myself later. Thanks for your great work!

ekqt:
Hey, @tmkx I'm trying to give this a test. I am able to find the documentation but I can't seem to find the package?

octref:
The package is marked as private now and will later be merged into renderer-svg. Maybe I should publish a temporary version...

PR #261, opened 2021-11-23
fix typo: grammer -> grammar

samuela:
- [n/a] Add a test if possible

- [n/a] This PR fixes #

- [n/a] I have read docs for [adding a theme](/docs/themes.md#adding-theme).

- [n/a] I have read docs for [adding a language](/docs/languages.md#adding-grammar).
- [n/a] I have searched around and this is the most up-to-date, actively maintained version of the language grammar.
- [n/a] I have added a sample file that includes a variety of language syntaxes and succinctly captures the idiosyncrasy of a language. See [docs](/docs/languages.md#adding-grammar) for requirement.

### DIFF
--- a/packages/shiki/src/registry.ts
+++ b/packages/shiki/src/registry.ts
@@ -46,7 +46,7 @@ export class Registry extends TextMateRegistry {
     return Object.keys(this._resolvedThemes) as Theme[]
   }
 
-  public getGrammer(name: string) {
+  public getGrammar(name: string) {
     return this._resolvedGrammars[name]
   }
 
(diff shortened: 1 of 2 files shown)


Gerrit0 (approved):

packages/shiki/src/registry.ts
```diff
@@ -46,7 +46,7 @@ export class Registry extends TextMateRegistry {
     return Object.keys(this._resolvedThemes) as Theme[]
   }
 
-  public getGrammer(name: string) {
+  public getGrammar(name: string) {
     return this._resolvedGrammars[name]
   }
```
orta, line 51: This is public API, so we'd need a `getGrammer` -> `getGrammar` so that we don't need to bump the major. Otherwise looks 👍🏻

Gerrit0: Shiki is still 0.x

orta:
Yeah, this API doesn't really seem to be used outside of shiki from the look of the repos too

samuela:
Thanks @orta and @Gerrit0 !

Issue #196, opened 2021-08-10
Error when handling Svelte code


markjaquith:
This rather standard Svelte code causes an error:

```js
const shiki = require('shiki')

shiki.getHighlighter({
    theme: 'nord'
})
.then(highlighter => {
    console.log(highlighter.codeToHtml(`<button on:click={increment}>Click</button>`, 'svelte'))
})
```

> `UnhandledPromiseRejectionWarning: Error: 105,110,118,97,108,105,100,32,112,97,116,116,101,114,110,32,105,110,32,108,111,111,107,45,98,101,104,105,110,100
    at OnigScanner.findNextMatchSync (/Users/mark/git/shiki-svelte-bug/node_modules/onigasm/lib/OnigScanner.js:100:27)
    at /Users/mark/git/shiki-svelte-bug/node_modules/vscode-textmate/release/main.js:1:9891
    at R (/Users/mark/git/shiki-svelte-bug/node_modules/vscode-textmate/release/main.js:1:10242)
    at f (/Users/mark/git/shiki-svelte-bug/node_modules/vscode-textmate/release/main.js:1:12144)
    at S (/Users/mark/git/shiki-svelte-bug/node_modules/vscode-textmate/release/main.js:1:11997)
    at w (/Users/mark/git/shiki-svelte-bug/node_modules/vscode-textmate/release/main.js:1:9336)
    at f (/Users/mark/git/shiki-svelte-bug/node_modules/vscode-textmate/release/main.js:1:13085)
    at S (/Users/mark/git/shiki-svelte-bug/node_modules/vscode-textmate/release/main.js:1:11997)
    at e._tokenize (/Users/mark/git/shiki-svelte-bug/node_modules/vscode-textmate/release/main.js:1:8354)
    at e.tokenizeLine2 (/Users/mark/git/shiki-svelte-bug/node_modules/vscode-textmate/release/main.js:1:7575)`

PuruVJ:
Facing exact same issue 👍

Valexr:
in this repo https://github.com/shikijs/shiki-playground shiki uses in frontend part...
set WASM & CND – https://github.com/shikijs/shiki-playground/blob/main/src/highlighter.ts
& initialize app after shiki  – https://github.com/shikijs/shiki-playground/blob/main/src/main.ts

bluwy:
This seems to be fixed in the latest version (0.10.0), not getting the errors anymore.

markjaquith:
[...]

PR #265, opened 2021-11-24
fix: Update oniguruma to version used by VSCode

### LINKED ISSUE #264: Shiki uses a different Oniguruma version than VSCode
The Svelte grammar used by Shiki tools includes the following lines ([source](https://github.com/sveltejs/language-tools/blob/6c5633a6b64ad995a4cf5bf08eca771d0a583669/packages/svelte-vscode/syntaxes/svelte.tmLanguage.src.yaml#L389-L390)):

```yaml
    - { match: '(?<=(use|transition|in|out|animate):).*$', name: variable.function.svelte  }
    - { match: '(?<=(let|class):).*$',                     name: variable.parameter.svelte }
```

This is a problem for Shiki since the version Oniguruma does not support variable length expressions in lookbehind groups. However, this works properly in VSCode.

image

I did... entirely too much digging, and eventually found that VSCode uses a release of oniguruma from October 2020, while Shiki uses a random commit from April 2018.

Cross linking:
- https://github.com/TypeStrong/typedoc/issues/1801

Gerrit0:
Resolves #264
Resolves TypeStrong/typedoc#1801

- [x] Add a test if possible
- [x] This PR fixes #264


Gerrit0:
This should also fix #196

orta:
Great, yeah, this makes sense 👍🏻

ryanflorence:
Looks like this also fixes some memory leaks we've been investigating with this package 🎉

PR #266, opened 2021-11-25
chore: Bump Shiki version to 0.9.13

Gerrit0:
Release includes:

- Updated grammars / themes
- fix: Update oniguruma to version used by VSCode 
- fix typo: grammer -> grammar (technically a breaking change, but it doesn't look like anyone uses it, so I left this as a patch release)


orta:
👍🏻

PR #290, opened 2022-01-20
Add bgMinWidth parameter to renderer-svg

### LINKED ISSUE #289: [renderer-svg] - feature request: add bgMinWidth parameter that will enforce minimal width of the generated svg
Hey shiki maintainers!
I wanted to say thank you for this awesome package and was thinking about improving it with new parameter in the options bag.
This is supposed to be an additive change to the options bag that will introduce `bgMinWidth` parameter that will enforce generated SVG minimal width. Logic is simple: if this parameter is not provided then logic doesn't change, if it's provided we pick highest number between width calculated by measurements and this parameter.
I already prepared PR for this.

TheFedaikin:
- [x] Add a test if possible
- [x] This PR closes #289  

Hey maintainers of shiki monorepo!

I also did some code cleanup stuff, which is non-breaking and mostly based around my very strict local ESLint config being absolutely mad about some of the code it encountered. Since I took some liberties with code "cleanup" I am willing to revert them if they don't sit right with you, because obviously this is not my repo, and I don't want to be enforcing stuff on people.

I am also willing to add tests for everything, but before that I'd like to know if I can keep the "code" cleanup stuff.


TheFedaikin:
I assume there's no interest in this, I'll close this one.

octref:
Oh no this is actually pretty good. Sorry I'm a bit busy with life. I will try to get this in sometime this weekend.

TheFedaikin:
Hey, thanks for the review, and dw about being busy - I closed this one so it wouldn't be pending, but if you are interested, that's great! 

I'll try to get some time to add tests in here as well.

octref:
Yeah you can use either Jest snapshots: https://jestjs.io/docs/snapshot-testing or use regex to extract the generated SVG width and do some assertions. Good luck!

TheFedaikin:
@octref  Sorry for the long wait, life's been crazy last months, and I totally got carried away from finishing this PR. I've added snapshot testing for a measurements reference (as well as some expects). I think this looks sufficient from my side, so I am looking towards a review. I wanted to do a small opportunistic refactoring again, but decided against it so this PR is not bloated with a lot of different stuff.

I'd like to discuss some additional features for the `svg-rendering` module (I use it heavily in my workflow, so I'd like to contribute to the greater community, but definitely don't want to overreach any boundaries). I assume it's best done through the issues since there's no discussions tab/discord link. Happy to start this conversation to allow people to do stuff like `https://carbon.now.sh/` but from cli side, if you are interested.

orta:
This looks structurally good to me, we're not breaking any tests and the stylistic changes all feel pretty good too. We should take it.

As you might have guessed, we treat shiki as a very mature project so things take a while. I'm open to us being faster in the svg rendering module as it's not really 'hot code path' for the same level of infra as the core shiki module.

PR #294, opened 2022-02-01
fix(docs): fixed custom language instructions

### LINKED ISSUE #280: Example of adding a language doesn't work
I have a tmlanguage converted to JSON and I am attempting to load it into a highlighter according to the docs, e.g.:

                langs: [langJson]
            });

            const code = highlighter.codeToHtml("sample language", { lang: 'mine', theme: 'dark-plus'});

This fails.  In the callstack, shiki is calling loadGrammar:

    async loadGrammar(scopeName) {
        const lang = this.scopeToLangMap[scopeName];
        if (!lang) {
            return null;
        }
        if (lang.grammar) {
            return lang.grammar;
        }
        const g = await fetchGrammar(languages.includes(lang) ? `${this.languagesPath}${lang.path}` : lang.path);
        lang.grammar = g;
        return g;
    }

lang resolves to:

{fileTypes: Array(1), name: 'mine', scopeName: 'source.mine', patterns: Array(38)}

lang.grammar is null, so then it attempts to call fetchGrammar.

languages.includes(lang) returns false, and lang.path is null, so fetchGrammar receives null as its parameter and fails to work.

This feels like a bug, but if it's a usage problem, how do I make this work?

I am not forking shiki or anything - I used npm and I am loading the tmlanguage json from an external source, not even a disk file.  The docs imply this should work.

Can anyone please help me?

JuanM04:
- [x] This PR fixes #280

### DIFF
--- a/docs/languages.md
+++ b/docs/languages.md
@@ -15,11 +15,17 @@ You can add a new language to your shiki instance by using the JSON version of y
 import shiki from "shiki"
 import {readFileSync} from "fs"
 
-const myLanguage = JSON.parse(readFileSync("./path/to/lang.tmLanguage.json"))
+const myLanguageGrammar = JSON.parse(readFileSync("./path/to/lang.tmLanguage.json"))
 
-shiki.getHighlighter({
-    langs: [...shiki.BUNDLED_LANGUAGES, myLanguage]
-})
+const myLanguage = {
+  id: "mylanguage",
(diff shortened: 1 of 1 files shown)


orta:
Cool, looks good to me - thanks!

EvanMachusakNCQA:
I tried this example and it works perfectly.  Thank you!

PR #298, opened 2022-02-09
Split plaintext code into lines

silvenon:
- [x] Add a test if possible

Currently it isn't possible to highlight lines in `txt` code blocks because the entire code is being treated as a single line. This changes makes the highlighter respect line breaks.


octref:
Good work, thanks!

PR #308, opened 2022-03-18
vscode-textmate@6.0.0

### LINKED ISSUE #307: Build fails on Linux ARM64 with yarn.
https://gist.github.com/jonshipman/f7f7919c3d2840538c30fe15015d3f04

See above gist. Using Yarn 3.2.0 and node-modules nodeLinker inside WSL Ubuntu 20.04, build fails. This is due to vscode-textmate being locked to v5.2.0 in https://github.com/shikijs/shiki/blob/main/packages/shiki/package.json. Changing the resolution in my package to 6.0.0 the install will complete.

jonshipman:
closes #307

### DIFF
--- a/packages/shiki/package.json
+++ b/packages/shiki/package.json
@@ -33,7 +33,7 @@
   "dependencies": {
     "jsonc-parser": "^3.0.0",
     "vscode-oniguruma": "^1.6.1",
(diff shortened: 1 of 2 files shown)


octref:
Thanks! Not sure if this will have any side effects but I guess we won't know until we make a release with it.

PR #322, opened 2022-06-14
fix: add 'yml' as an alias for yaml

qwqcode:
- [x] ~~Add a test if possible~~


orta:
Thanks!

PR #325, opened 2022-06-21
doc: update `HighlighterOptions.langs` JSDoc

### LINKED ISSUE #324: Incorrect `HighlighterOptions.langs` jsdoc
The `HighlighterOptions.langs` jsdoc indicates that by default it would only load upfront the `html`, `css` and `js` language definitions. 

https://github.com/shikijs/shiki/blob/b778187e978b24be50e9900d82e02e6afaeeb185/packages/shiki/src/types.ts#L19-L24

However at runtime, it appears that all the languages are loaded upfront by default: [live example](https://stackblitz.com/edit/node-ssworr?file=index.js)

```js
import { getHighlighter } from 'shiki';

const highlighter = await getHighlighter({});
console.log(highlighter.getLoadedLanguages().length);
// Expected: 3
// Actual: 164
```

pmdartus:
- [ ] Add a test if possible
- [X] Format all commit messages with [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/)
- [X] This PR fixes #324

Update the JSDoc comment to reflect the actual runtime behavior.

### DIFF
--- a/packages/shiki/src/types.ts
+++ b/packages/shiki/src/types.ts
@@ -19,7 +19,7 @@ export interface HighlighterOptions {
   /**
    * A list of languages to load upfront.
    *
-   * Default to `['html', 'css', 'javascript']`
+   * Default to all the bundled languages.
    */
   langs?: (Lang | ILanguageRegistration)[]
 


orta:
Thanks

PR #327, opened 2022-06-21
fix: Don't preload any lang when option is set

### LINKED ISSUE #326: Capability to create a new `Highlighter` without loading any language upfront
## Description

There is currently no way to create a new `Highlighter` without loading any language upfront. Setting the `HighlighterOptions.langs` to an empty array results in loading all the bundled languages.

## Example

```js
import { getHighlighter } from 'shiki';

const highlighter = await getHighlighter({ langs: [] });
console.log(highlighter.getLoadedLanguages().length);
```

### Actual behavior

```
// Print: 164
```

### Expected behavior

```
// Print: 0
```

## Workaround

The current workaround is to add a single language to the `langs` options to force Shiki to only load this language instead of preloading all the languages upfront.

pmdartus:
- [X] Add a test if possible
- [X] Format all commit messages with [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/)
- [X] This PR fixes #326

### DIFF
--- a/packages/shiki/src/__tests__/issues.test.ts
+++ b/packages/shiki/src/__tests__/issues.test.ts
@@ -25,3 +25,12 @@ test('Correctly handles variable length lookbehind (#264)', async () => {
   })
   expect(out).toMatchSnapshot()
 })
+
+// https://github.com/shikijs/shiki/issues/326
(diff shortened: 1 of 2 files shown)


orta:
Agree 👍🏻

PR #333, opened 2022-07-21
feat: don't throw Error 'No language registration for xxx'

### LINKED ISSUE #332: don't throw Error 'No language registration for xxx'
When I call getGrammar('mjs')  in `packages\shiki\src\highlighter.ts`, error occurs: `No language registration for mjs`, which results in a error like ['vitepress build docs' error](https://github.com/vuejs/vitepress/issues/1025#issuecomment-1191481726).

As a result, I **pass the default language 'text'** to `function codeToThemedTokens()` when there is no language registration. This will solve the problem.

gittig11:
- [✓] This PR fixes #332

### DIFF
--- a/packages/shiki/src/highlighter.ts
+++ b/packages/shiki/src/highlighter.ts
@@ -116,6 +116,10 @@ export async function getHighlighter(options: HighlighterOptions): Promise<Highl
     theme?: IThemeRegistration,
     options = { includeExplanation: true }
   ) {
+    let _agrammar = _registry.getGrammar(lang)
+    if (!_agrammar) {
+      lang = 'text'
+    }
     if (isPlaintext(lang)) {
       const lines = code.split(/\r\n|\r|\n/)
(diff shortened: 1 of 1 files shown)


Gerrit0:
This seems like a bad idea to me. Shiki shouldn't just silently not highlight code.

orta:
Yep, we add support for new languages when people send PRs adding them 👍🏻 - I agree with @Gerrit0 that we probably shouldn't fallback in shiki to safely NOOPing. 

This leaves the choice of what to do when that language isn't there to the integrators like vitepress in your case instead of silently giving a false positive.

gittig11:
Maybe both change `throw Error()` to just `console.log('No language registration for xxx')` and pass the default language 'text' to codeToThemedTokens() when facing with unsupported languages would be a better choice. ==

orta:
Thanks for giving it a shot! I'm going to close this PR to keep the open ones actionable

PR #335, opened 2022-08-11
Fix CI

orta:
The SVG rendering PR improvements in #290 ( /cc @TheFedaikin ) are OS dependent and can't be used on our CI ( for example on my computer (linux) :

```diff
- <svg viewBox=\\"0 0 156.4921875 101.9921875\\" width=\\"156.4921875\\" 
+ <svg viewBox=\\"0 0 156.4921875 112\\" width=\\"156.4921875\\" height=\\"112\
```

So, I've amended the tests to not verify the full snapshots and only to measure the viewBox width - which is _kinda_ the key of the test. If you're up for it, I'll happily take PRs adding what you want to those tests but I'm cool with these changes

Also updates the pnpm lockfile because it did it


orta:
Looks like there was some more, I think to safely render these snapshots across windows and linux etc we might need a clever ish regex to remove the width=x and height=x from the snapshots?

PR #358, opened 2022-09-13
fix: Use WASM in non-browser environments

jeffrey-ng:

### DIFF
--- a/packages/shiki/src/loader.ts
+++ b/packages/shiki/src/loader.ts
@@ -15,7 +15,7 @@ export const isBrowser = isWebWorker || !isNode
 
 // to be replaced by rollup
 let CDN_ROOT = '__CDN_ROOT__'
-let WASM: string | ArrayBuffer = ''
+let WASM: string | ArrayBuffer | undefined = undefined
 
 /**
  * Set the route for loading the assets
@@ -46,7 +46,7 @@ export async function getOniguruma(): Promise<IOnigLib> {
   if (!_onigurumaPromise) {
     let loader: Promise<void>
 
-    if (isBrowser) {
+    if (isBrowser || WASM) {
       if (typeof WASM === 'string') {
         loader = loadWASM({
           data: await fetch(_resolvePath('dist/onig.wasm')).then(r => r.arrayBuffer())


orta:
I think this PR probably needs a bit of context (as there's no PR body) what are your goals?

pengx17:
I too encounter a problem when integrating shika into the new [App directory feature](https://beta.nextjs.org/docs/routing/fundamentals) in Next.js 13. 
It seems that when invoking `shiki.getHighlighter` in a [`generateStaticParams` call](https://beta.nextjs.org/docs/api-reference/generate-static-params), it is actually running in the Webpack context that has a custom `require`. See a similar issue here https://github.com/webpack/webpack/issues/9393. 

I have no clue why Nextjs is doing this, but I suppose the Nextjs team wants to have the server code bundled as well so that the runtime on Edge could be minimized/optimized.

ishaba:
Looks like i also will fix my issue with nuxt any updates on this?

octref:
Is this still an issue, @pengx17?

pengx17:
> Is this still an issue, @pengx17?

I just [reworked](https://github.com/pengx17/next-blog-neo/commit/fca039227ec821fe25a7ea5898e13e0536877aa6#commitcomment-98383609) my blog and rendering code blocks directly inside of React server components. Shiki works perfectly in this case. 👯 
I guess my original issue was that I should not call shiki inside of [generateStaticParams](https://beta.nextjs.org/docs/api-reference/generate-static-params). I am not too sure if it was still an issue.

Thanks for providing this awesome project!

octref:
Great, I'll close this then. There's a sample for next + shiki now: https://github.com/shikijs/next-shiki

samselikoff:
@octref just came across this issue! I tried the `next-shiki` sample you posted using Vercel's `experimental-edge` runtime and I'm hitting a similar wasm error:

image

I'm a bit out of my league here as I haven't spent a lot of time on this but just curious if shiki should be able to work in the edge-runtime? Thanks for any pointers!

pengx17:
@samselikoff I think the reason of the error is that non-module (json here) in `node_modules` will not be bundled into the edge runtime. I am not sure if there is a graceful way to fix this issue, but my solution is to copy/paste shiki resource files (/themes, /lanague) under the next repo, and then configure shiki to point to the files in the repo instead of using the default path:
See https://github.com/pengx17/next-blog-neo/blob/main/app/lib/rehype-shiki.ts#L16-L17

PR #368, opened 2022-10-24
fix: increase contrast for reds on the light plus theme

### LINKED ISSUE #366: Request for a light_vs.json update
We just recently fixed a color contrast issue for code.visualstudio.com HTML and CSS code blocks by making some foreground red colors a little darker.

VS Code commit https://github.com/microsoft/vscode/commit/bf220dd91e46d061cf37dbb6e0092370519ab8e1

We'd like a shiki update so that this change can be reflected on the VS Code website. Not sure the cadence for shiki updates or what is the best process for creating an update request.

Thanks!

rzhao271:
- [ ] Add a test if possible

- [x] This PR fixes #366

### DIFF
--- a/packages/shiki/themes/light-plus.json
+++ b/packages/shiki/themes/light-plus.json
@@ -76,7 +76,7 @@
     {
       "scope": "entity.other.attribute-name",
       "settings": {
-        "foreground": "#ff0000"
+        "foreground": "#e50000"
       }
     },
     {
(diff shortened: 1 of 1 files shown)


octref:
Hi Raymond! There's an [Action](https://github.com/shikijs/shiki/actions/workflows/update-themes.yml) that automatically fetches latest theme from vscode. In the main branch I already see this change reflected. Thanks!

PR #361, opened 2022-09-21
feat: upgrade to vscode-textmate-v7. fix #343

octref:
Still need to fix the failing tests.

PR #377, opened 2022-11-27
feat: dependency updates incl. textmate

### LINKED ISSUE #343: Missing italic styles in 0.11?
Reproduction (https://runkit.com/brc-dd/6309d1529e302e000a926cad):

```js
const shiki_10 = require('shiki@0.10.1')
const shiki_11 = require('shiki')

const h_10 = await shiki_10.getHighlighter({ theme: 'material-palenight' })
const h_11 = await shiki_11.getHighlighter({ theme: 'material-palenight' })

console.log(h_10.codeToHtml(`import foo from 'bar'`, { lang: 'js' }))
console.log(h_11.codeToHtml(`import foo from 'bar'`, { lang: 'js' }))
```

Output (notice the difference in style of `import` and `from`):

image

Is some additional config required to render it like before, or the theme itself has changed?

muenzpraeger:
- [x] Add a test if possible

- [x] This PR fixes #343, #361 

This PR updates all dependencies to recent versions (which actually also caused a few other changes). The change also included the update to the most recent `vscode-textmate` version.


jest.config.js
```diff
@@ -4,12 +4,19 @@ module.exports = {
   testTimeout: 10000,
   testEnvironment: 'node',
   globals: {
-    __BROWSER__: false,
-    'ts-jest': {
-      tsconfig: {
+    __BROWSER__: false
+  },
+  snapshotFormat: {
+    escapeString: true,
```
muenzpraeger, line 10: There was a change in Jest -> they don't escape anymore. For the sake of simplicity (no updated snapshots) in this PR I set the tests to use the previous behaviour.

jest.config.js
```diff
@@ -4,12 +4,19 @@ module.exports = {
...
+    printBasicPrototype: true
+  },
+  transform: {
+    '^.+\\.tsx?$': [
```
muenzpraeger, line 14: Another Jest change, this needs now to be a part of the `transform` object.

packages/renderer-svg/src/__tests__/styleAttributes.test.ts
```diff
@@ -42,6 +42,5 @@ test(`SVG renderer should ignore minimal background width when it's own boundari
   })
 
   const out = (await r).renderToSVG([[{ content: 'foo', color: '#aabbccff' }]])
-  expect(out).toContain(`<svg viewBox="0 0 150`)
-  // expect(out).toMatchSnapshot()
+  expect(out).not.toContain(`<svg viewBox="0 0 150 `)
```
muenzpraeger, line 45: The test was failing b/c of a wrong check. ;-)

packages/shiki/src/stackElementMetadata.ts
```diff
@@ -135,6 +146,7 @@ export class StackElementMetadata {
       ((_languageId << MetadataConsts.LANGUAGEID_OFFSET) |
         (_tokenType << MetadataConsts.TOKEN_TYPE_OFFSET) |
         (_fontStyle << MetadataConsts.FONT_STYLE_OFFSET) |
+        (_containsBalancedBracketsBit << MetadataConsts.BALANCED_BRACKETS_OFFSET) |
```
muenzpraeger, line 149: Only implemented, but not really exposed (yet).

packages/shiki/package.json
```diff
@@ -5,6 +5,7 @@
   "author": "Pine Wu <octref@gmail.com>",
   "homepage": "https://github.com/octref/shiki/tree/main/packages/shiki",
   "license": "MIT",
+  "type": "module",
```
drwpow, line 8: This will break the package because `dist/index.js` is still CommonJS. This should be removed unless the `main` import is also changed to `dist/index.esm.js`, or an [import map with conditional exports](https://nodejs.org/api/packages.html#conditional-exports) is also added (which IMO is a breaking change and should be its own PR)

packages/shiki/package.json
```diff
@@ -33,7 +34,7 @@
   "dependencies": {
     "jsonc-parser": "^3.0.0",
     "vscode-oniguruma": "^1.6.1",
```
drwpow, line 36: This can be bumped as well

drwpow:
This PR would help the error I’m currently encountering (Node 18, macOS 13.x, Apple silicon chip):

```
import object field 'setTempRet0' is not a Function
```

Bumping `vscode-oniguruma` and `vscode-textmate` to the most recent versions seems to fix that error.

packages/shiki/package.json
Gerrit0: Yes, please don't do this. TypeDoc depends on Shiki, and I *really* don't want to deal with the pain that is ESM.

packages/renderer-svg/src/__tests__/styleAttributes.test.ts
Gerrit0: I'm not convinced the check was wrong... I think something updated in the renderer which resulted in an output change, but checking not contains is very suspicious...

packages/shiki/rollup.config.js
```diff
@@ -118,7 +121,7 @@ export default defineConfig([
     plugins: [
       dts(),
       copy({
-        targets: [{ src: require.resolve('vscode-oniguruma/release/onig.wasm'), dest: 'dist' }]
+        targets: [{ src: resolve('node_modules/vscode-oniguruma/release/onig.wasm'), dest: 'dist' }]
```
Gerrit0, line 124: Why?

Gerrit0 (changes requested):

packages/renderer-svg/src/__tests__/styleAttributes.test.ts
muenzpraeger: I _think_ the change is correct. Here's why:

- The test description says it is a check that the render should _ignore_ the minimal background width.
- The [test passes](https://github.com/shikijs/shiki/blob/main/packages/renderer-svg/src/__tests__/styleAttributes.test.ts#L41) a value of 150 to `getSVGRenderer` for `bgMinWidth`. So that's what the SVG should have at minimum. And that's the value that should be ignore - according to the test description - if  the SVG width is higher.
- The renderer [validates via Math.max() the actual width against the min width](https://github.com/shikijs/shiki/blob/main/packages/renderer-svg/src/index.ts#L170-L173). In this test case the actual width is higher than the min width - so the actual width will be used instead of the min width.
- If the min width would be set on the SVG, then `<svg viewBox="0 0 150 ` would be correct. But as the actual width should override the min width the test should only pass when it's _not_ `<svg viewBox="0 0 150 `.

Happy to be proven wrong though, first time for me in that codebase. 🤷🏼‍♂️

packages/shiki/rollup.config.js
muenzpraeger: The usage of `require.resolve` [is throwing ](https://nodejs.org/docs/latest-v18.x/api/esm.html#esm_no_require_resolve) b/c of the module nature of Rollup.

The alternative option to get back `require.resolve` (and not to use `path.resolve`) would be to re-create require:

```javascript
import { createRequire } from 'node:module';
const require = createRequire(import.meta.url);
```

I've no strong opinion on either of both approaches.

packages/shiki/package.json
muenzpraeger: Good point, oversight on my end. Removed this, and instead renamed the rollup config files (which actually have been the driver for this).

packages/renderer-svg/src/__tests__/styleAttributes.test.ts
Gerrit0: Looks like you're correct, thanks! I missed the "not" in the test description

muenzpraeger: Been there too. ;-)

packages/shiki/rollup.config.js
Gerrit0: Ah, makes sense! I didn't remember the module change when seeing this. This is probably fine, assuming hoisting doesn't change the actual file location

jest.config.js
orta: I made those changes in Jest ;) - but I dig not making things change unnecessarily

orta:
Yeah, I agree, this is a good PR 

I'll try commit some time this weekend to deploying a new shiki release 👍🏻

muenzpraeger:
@orta Cool! I'll have 2-3 more PRs in pipe (think of them as proposals, so not mandatory if you disagree). I'll try to get them in within the next 24h.

One of them is actually also adding more docs. ;-)

muenzpraeger:
With the latest commit we can eventually supersede #379 (added embedded language resolution to IGrammarConfiguration)

orta:
OK, that's  green now - biggest question mark on the PR is the removal of `IRawThemeSetting` to `any` but it doesn't seem to be user-facing so it's not a blocker IMO

Thanks for this @muenzpraeger - lets see if I can get all this wrangled into a release after I go through the other PRs.

muenzpraeger:
Also not a big fan of the `any` changes either, but I figured they don't have an impact on the implementation.

I've seen on the [vscode-textmate repo this issue](https://github.com/microsoft/vscode-textmate/issues/185) with no response yet, so potentially an oversight on their end.

orta:
NP, all these changes just got shipped in 0.12.0 - thanks!

PR #395, opened 2022-12-20
fix: minor doc updates

muenzpraeger:
Minor docs update

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,6 @@ And after that you can (and should) check out the reference docs for deeper dive
 
 - [Themes](./docs/themes.md)
 - [Languages](./docs/languages.md)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -326,7 +326,7 @@ import { getHighlighter, setWasm } from 'shiki'
 
 // It is recommended to use a Response object. Oniguruma will then use WebAssembly.instantiateStreaming(), which
 // means it can start parsing the file while it's still downloading.
-const wasmResponse = new Response(await fetch('/your/path/onig.wasm'))
+const wasmResponse = await fetch('/your/path/onig.wasm')
```
orta, line 329: Wouldn't this await force the full load ahead of time, instead of streaming like the comment? (both old and new may be wrong here?)

PR #402, opened 2023-01-07
fix: Type errors in Node projects without DOM lib

### LINKED ISSUE #401: Type errors due to assuming DOM libraries in Node projects
I upgraded TypeDoc to use the latest Shiki, and started seeing this error, since it does not include "DOM" in its lib and skipLibCheck is off.

> Error: node_modules/shiki/dist/index.d.ts(316,55): error TS2304: Cannot find name 'Response'.

Gerrit0:
- [x] This PR fixes #401


orta:
I think the test fails are legit - they probably are running typescript inline

Gerrit0:
... jest is weird. Made it work, there's probably a better way of making `__BROWSER__` visible to the tests

PR #408, opened 2023-01-11
 doc: add pointer to reverse proxe MIME config

### LINKED ISSUE #400: wasm file streaming issue
Updating to 12 introduced an error for me and I wanted to share it in case others encounter it. I'm not sure how widespread this occurs. 

> Failed to execute 'compile' on 'WebAssembly': Incorrect response MIME type. Expected 'application/wasm'.

The change is from using a Response object ([here](https://github.com/shikijs/shiki/blob/main/packages/shiki/src/loader.ts#L18)) instead of an ArrayBuffer, which as you know is handled differently by [oniguruma](https://github.com/microsoft/vscode-oniguruma/blob/main/out/index.js#L337)

Using the advice in the readme works:

```js
const wasmBuffer = await fetch('/your/path/onig.wasm').then(res => res.arrayBuffer())
setWasm(wasmBuffer)
```

muenzpraeger:
Documentation update for specifying MIME types on reverse proxies

Fixes #400

### DIFF
--- a/README.md
+++ b/README.md
@@ -293,6 +293,8 @@ For Node.js environments the WASM file is automatically loaded, for browser envi
 
 > If used, `setWASM` must be invoked before `getHighlighter` is called.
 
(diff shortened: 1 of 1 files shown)


orta:
👍🏻

PR #415, opened 2023-01-15
doc: Fix a way to access `renderToHTML`

mactkg:
Thanks for creating such a beautiful syntax highlighter!

I try using an example code of customizing code block rendering on README, but it doesn't work. It seems only [able to access renderToHTML from `shiki` global object](https://github.com/shikijs/shiki/blob/main/packages/shiki/src/index.ts#L9), [not a highlighter](https://github.com/shikijs/shiki/blob/main/packages/shiki/src/highlighter.ts#L204-L217). I fixed the doc.

## checklists
- [x] Add a test if possible

### DIFF
--- a/README.md
+++ b/README.md
@@ -365,7 +365,7 @@ If you want to render the tokens into a code yourself, Shiki exposes two key met
 - `renderToHTML` takes an array of tokens and returns an HTML string that represents the provided code.
 
 ```js
-import { getHighlighter } from 'shiki'
+import shiki, { getHighlighter } from 'shiki'
 
 const highlighter = await getHighlighter({
   theme: 'nord',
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -365,7 +365,7 @@ If you want to render the tokens into a code yourself, Shiki exposes two key met
 - `renderToHTML` takes an array of tokens and returns an HTML string that represents the provided code.
 
 ```js
-import { getHighlighter } from 'shiki'
+import shiki, { getHighlighter } from 'shiki'
```
mactkg, line 368: I didn't import `renderToHTML` using named imports in a sample code because the second code block omits import statements. I think that it could be confusing whether `renderToHTML` is a global object.

muenzpraeger: Good catch on the documentation.

Although I'm not sure if we should advocate for _not_ using named imports, that seems counter intuitive to me.

I wonder if we should look at the overall API design for `renderToHTML`, and add it to the highlighter. @orta or @octref may give some color why it's historically not part of the highlighter.

octref: Yeah this is a bad API. Will be cleaner for 1.0. Let's fix the docs for now so it's consistent between code/doc.

octref (approved):

octref:
Thanks for your contribution 🚀

PR #422, opened 2023-01-26
feat: add powerquery and dax grammars

### LINKED ISSUE #421: PowerQuery and DAX
Would you accept a PR to include PowerQuery and Dax languages?

Here's PowerQuery: https://github.com/microsoft/powerquery-language

Maybe this is DAX? Does this look right? https://github.com/huyza/dax-language/blob/master/syntaxes/dax.grammer.json

KevinBatdorf:
- [x] Add a test if possible

- [x] This PR fixes #421 

Dax:

Screen Shot 2023-01-25 at 11 35 49 PM

PowerQuery
Screen Shot 2023-01-25 at 11 41 04 PM


scripts/grammarSources.ts
```diff
@@ -316,6 +318,7 @@ export const languageAliases = {
   markdown: ['md'],
   'objective-c': ['objc'],
   powershell: ['ps', 'ps1'],
+  powerquery: ['m'],
```
octref, line 321: Overall good – why adding this alias though? Seems too easy to confuse with other languages. For example objc can also have `.m` files.

octref (changes requested):

packages/shiki/src/languages.ts
```diff
@@ -634,6 +642,7 @@ export const languages: ILanguageRegistration[] = [
     id: 'marko',
     scopeName: 'text.marko',
     path: 'marko.tmLanguage.json',
+    samplePath: 'marko.sample',
```
octref, line 645: Good catch thanks!

scripts/grammarSources.ts
KevinBatdorf: The language is new to me so I could be wrong buut it appears to be an alias for it.

https://learn.microsoft.com/en-us/power-query/power-query-what-is-power-query#power-query-m-formula-language

But I think it can be left off until someone requests it. I can update the PR tomorrow.

packages/shiki/src/languages.ts
KevinBatdorf: Ah I actually meant to remove this change. There were a few errors when building about missing grammar files and other things. I was going to submit another PR to cover those but got called away. I can do that tomorrow though.

scripts/grammarSources.ts
orta: Aye, would recommend removing - `.m` is the default for objective-c which would conflict with this I imagine

KevinBatdorf: I've removed it now. I'm going to submit on another PR with some build issues I ran into in a bit. This one is ready to be merged in once approved.

octref (approved):

octref:
Looks great now, thanks @KevinBatdorf 🚀

PR #429, opened 2023-01-31
fix: add tabindex of 0 to default pre element

### LINKED ISSUE #428: Scrollable <pre> elements should be focusable for WCAG 2.1 compliance
As mentioned in https://github.com/shikijs/shiki/issues/220, scrollable `<pre>` elements should have `tabindex="0"` to allow them to be focusable.  Right now, these elements will fail WCAG audits if/when they become scrollable. https://www.w3.org/WAI/standards-guidelines/act/rules/0ssw9k/proposed/

It seems ideal to me to make this the default rather than try to update consumers like `remark-shiki`, etc. to pass in a custom `pre`.

I wouldn't expect this to be a breaking change since anyone who is already providing a `tabindex` will already be passing in an entirely custom `pre` template literal that will override this default.

kevinleedrum:
- [x] Add a test if possible _(updated snapshots)_
- [x] This PR fixes #428 

Thanks!


octref (approved):
LGTM, thanks!

PR #431, opened 2023-02-03
feat: add kusto grammar

KevinBatdorf:
This add grammar for Kusto ([link](https://learn.microsoft.com/en-us/azure/data-explorer/kusto/query/))

I added the alias `kql` but I'm not 100% sure what effect aliases have so just making a note here.


octref:
It means `kql` language will use kusto grammar for rendering. Thanks for yoru contribution!

PR #442, opened 2023-03-07
docs: fix grammar of "grammar's"

comex:
It shouldn't have an apostrophe.


orta:
That CI fail is likely due to bots sending package updates - thanks for this!

PR #443, opened 2023-03-07
fix: faster parsing time for assets

### LINKED ISSUE #439: Question about `_fetchJSONAssets`
Hello again,

I'm wondering why `_fetchJSONAssets` uses `jsonc-parser` instead of `JSON.parse`. It looks like all of the bundled themes and languages are valid JSON:
```
find ./packages/shiki/ -name "*.json" | \
  grep -e "\(languages\|themes\)" | \
  xargs -n1 -I{} node -e "JSON.parse(fs.readFileSync('{}','utf-8'))"
```

`_fetchJSONAssets` takes around 300ms right now, but switching it to `JSON.parse` brings it down to 30ms. I ask because this is the source of a noticeable delay in start up time for https://github.com/banga/git-split-diffs.

banga:
This improves the parsing time for themes and grammars by first using `JSON.parse` and then falling back to `jsonc-parser`. Since all of the themes and grammars currently checked in are valid JSON, this speeds up startup time considerably:

```
import shiki from './packages/shiki/dist/index.js'
import { hrtime } from 'process';

const start = hrtime()
const highlighter = await shiki.getHighlighter({ theme: 'nord'});
console.log(hrtime(start)[1] / 1e6)
```

This goes from ~420ms to ~185ms on my machine.

This PR fixes #439

### DIFF
--- a/packages/shiki/src/loader.ts
+++ b/packages/shiki/src/loader.ts
@@ -121,15 +121,21 @@ async function _fetchAssets(filepath: string): Promise<string> {
 
 async function _fetchJSONAssets(filepath: string) {
   const errors: ParseError[] = []
-  const rawTheme = parse(await _fetchAssets(filepath), errors, {
-    allowTrailingComma: true
-  })
+  const assetString = await _fetchAssets(filepath)
+  let rawAsset
+  try {
+    // Try to parse as JSON first, since it's much faster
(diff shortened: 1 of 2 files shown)


supun-io:
Waiting for this to go live. This is the main issue we have with Shiki. Thanks!

PR #445, opened 2023-03-19
Update Kotlin grammar and add a sample

alllex:
- [ ] Add a test if possible

---

There was already a link to the Kotlin grammar. However, it was updated 4+ years ago and does not support highlighting some Kotlin features, such as string interpolation.


muenzpraeger:
Thanks for the contribution! Can you check if using the grammar used in the Kotlin LSP offers the same fidelity?

https://github.com/fwcd/kotlin-language-server/tree/main/grammars

octref:
@alllex It seems https://github.com/fwcd/kotlin-language-server/tree/main/grammars is more recently updated. Can you check if that fixes your issue and switch to that as source, if possible?

alllex:
I checked the grammar you suggested, and it seems to be working as expected. I updated the PR to point to that grammar for `kotlin` language.

Issue #454, opened 2023-03-29
Global `fetch` type is getting modified by shiki


Ayc0:
Having this code in shiki:

```ts
declare global {
  function fetch(url: string): Promise<Response>
}
```

also impacts the global `fetch` when using shiki, and this type isn't compatible with the native one provided by TS

orta:
Yeah, makes sense, I think the answer is to probably switch these

https://github.com/shikijs/shiki/blob/99a9a3dcd2f2504d36b0672968fb05be53fd73cb/packages/shiki/src/loader.ts#L7-L19

to be inline `declare xyc` in this file, which should make them not afffect the global scopes

Ayc0:
@orta I can try to draft a PR if you want

orta:
Give it a shot!

Ayc0:
@orta the PR is ready: https://github.com/shikijs/shiki/pull/457

octref:
Fixed by #457. Thanks for your contribution!

klippx:
No, it is even worse now. Please reopen the issue! 

Before I added `nextra` as a dependency to my project I had  `await result.text()` returning typed as `string` automatically by typescript lib dom:

```ts
interface Body {
    /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/Request/body) */
    readonly body: ReadableStream<Uint8Array> | null;
    /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/Request/bodyUsed) */
    readonly bodyUsed: boolean;
    /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/Request/arrayBuffer) */
    arrayBuffer(): Promise<ArrayBuffer>;
    /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/Request/blob) */
    blob(): Promise<Blob>;
    /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/Request/formData) */
    formData(): Promise<FormData>;
    /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/Request/json) */
    json(): Promise<any>;
    /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/Request/text) */
    text(): Promise<string>;
}
```

However, now this library got sucked in to the mix, it has ruined the `Response` typings:

```
declare global {
    interface Response {
        json(): Promise<any>;
        text(): Promise<any>;
    }
}
```

[...]

PR #457, opened 2023-03-31
fix: avoid poisoning global scope

Ayc0:
- [ ] Add a test if possible
   I didn't add any automated test, but I checked that after running `pnpm build`, there was no mention of `Window` in the `index.d.ts` file:
  image
   I didn't add any test because:
   -  this test would require to be ran **after** `pnpm build`
   - if we wanted to write to test to ensure that we aren't poisoning the global scope, we need to do something like "is Response, or Window, etc., in the generated code?" But `Response` is used so we cannot use that, and other may be used in the future. So I don't know which test we could write that could be future proof for this.
- [x] This PR fixes https://github.com/shikijs/shiki/issues/454

I couldn't use just `declare var fetch` and others, because we need to add to the global scope `Window` and others as they are used by shiki's node_modules (and so when running tests locally, we need them, see https://github.com/shikijs/shiki/actions/runs/4572982584/jobs/8072860830).

So I moved those to a `global.d.ts` so that they'd be present in the project, but not in the built code.

### DIFF
--- a/packages/shiki/src/global.d.ts
+++ b/packages/shiki/src/global.d.ts
@@ -1,1 +1,20 @@
-declare var __BROWSER__: boolean
+declare global {
+  var __BROWSER__: boolean
+
+  // These will not be defined for Node which don't have "DOM" in their lib.
+  // So declare the minimal interface we need.
+  // They are defined here and not in `loader.ts` to avoid getting injected in the built code.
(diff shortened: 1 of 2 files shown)


orta:
Nice work!

Ayc0:
Thanks!

PR #476, opened 2023-05-25
feat(theme): allow to disable default theme

### LINKED ISSUE #473: Remove the default loading of `nord` theme when `getHighlighter`
If `theme` or `themes` is empty when using `getHighlighter`, then the `nord` theme will be loaded by default. This results in an unnecessary theme load for situations where the theme needs to be loaded dynamically (or an unnecessary theme resource request in the browser environment).
For example:

```ts
// Here there will be an unwanted `nord` theme load.
// Because later the required theme will be loaded dynamically depending on the situation.
const highlighter = await shiki.getHighlighter({})

export async function highlight(code, lang, theme) {
  // Here the real theme will be loaded, e.g. `github-dark`
  await highlighter.loadTheme(theme)
  return highlighter.codeToHtml(code, { lang, theme })
}
```

Can we remove the default theme logic here? Or a way to disable it in the above situations. Thanks :)

https://github.com/shikijs/shiki/blob/b8e88e68d6e1923d57702268deaf922143abd55d/packages/shiki/src/highlighter.ts#L43-L45

(Willing to contribute a PR for this)

kricsleo:
- [ ] Add a test if possible

- [x] This PR fixes #473

I have also added some documentation notes, so please let me know if there are any changes that are not appropriate :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,9 +142,33 @@ Load the highlighter with a default theme and no languages:
 ```js
 // ONLY the languages specified will be loaded. It's an empty array, so Shiki won't  preload anything.
 // For every language that you need later you'll have to load it separately (see down below).
+// The default theme `nord` will be loaded when `themes` is an empty array, or neither `theme` nor `themes` is specified.
 const highlighter = await getHighlighter({
   theme: 'nord',
   langs: []
+
+  // Or 👇
+  // langs: []
+
+  // Or 👇
+  // themes: [],
+  // langs: []
+})
+```
+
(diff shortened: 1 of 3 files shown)


kricsleo:
Why CI failed？

octref:
I really don't like the magical behavior of an empty array or writing new docs by making the `readme.md` longer. I'm working on 1.0 and let me fix that there. Thanks for the contribution still! Sorry for the CI, it should be fixed now.

PR #384, opened 2022-12-08
feat: add documentation / optimize WASM loading

muenzpraeger:
- [x] Add a test if possible
- [x] Format all commit messages with [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/)

- [x] This PR fixes #382, helps #365 (potentially), fixes #347, fixes #216, fixes #246, fixes #346, fixes #221, fixes #317, fixes #206

This PR refines WASM loading, so that there is less setup needed. It also adds.... *drumroll* documentation. ;-)

PR #504, opened 2023-07-29
fix: Compiler errors in Node

Gerrit0:
TypeDoc doesn't have "DOM" in the lib, so doesn't have a global `Response` interface like DOM types do. The changes in #384 (cc @orta) broke me because it references that type now. This moves the interface into loader.ts so that it gets bundled in the declaration file output.

### DIFF
--- a/packages/shiki/src/global.d.ts
+++ b/packages/shiki/src/global.d.ts
@@ -10,10 +10,6 @@ declare global {
 
   var self: Window & typeof globalThis
   function fetch(url: string): Promise<Response>
-  interface Response {
-    json(): Promise<any>
-    text(): Promise<any>
-  }
 }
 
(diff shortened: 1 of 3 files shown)


Gerrit0:
Deploy failures seems unrelated, I didn't change anything to do with node versions.

orta:
I've not thought too hard about this, but we did recently change this: https://github.com/shikijs/shiki/pull/457

Gerrit0:
I could certainly change it so that this is a local type that is exported, and name it ResponseLike or something...  so long as:

1. I don't need to add DOM to my lib
2. I can compile without skipLibCheck

Then I'm happy

octref (approved):
Think that's good for now. In the new API I'm working on I'll avoid this.

PR #511, opened 2023-08-16
feat: add additional sample code

KevinBatdorf:
This PR 

1. adds the missing sample codes (I believe I covered them all here)
2. renames `Marko.sample` to `marko.sample` to reflect the manifest file.
3. Updates a few that didn't seem like good representations of the language (too short, not enough tokens, etc)

I tried to find reputable sources for all the code.


orta:
Wow, that must have been quite some work, we should get this in regardless of those missing snapshots

KevinBatdorf:
Was quite tedious but needed them for a side project anyway 😅

PR #519, opened 2023-09-07
fix: alias `kt` to `kotlin`

scarf005:


antfu:
Sorry I went into #525 first. Thanks!

PR #523, opened 2023-09-22
chore: Update Dart grammar for change from inline classes to extension types

parlough:
Updates to the latest release of the official Dart syntax: https://github.com/dart-lang/dart-syntax-highlight/blob/master/grammars/dart.json


antfu:
We use scripts to update grammar automatically, and it should be already been updated now. Thanks.

parlough:
I didn't realize. Thanks for the clarification (and the update!).

PR #531, opened 2023-10-12
fix: npm publish only `dist` files

styfle:
This PR adds the `files` allowlist to package.json which will prevent assets like `languages` and `themes` from being included twice.

For example:
- https://unpkg.com/browse/shiki@0.14.5/languages/
- https://unpkg.com/browse/shiki@0.14.5/dist/languages/

Learn more: 
- https://docs.npmjs.com/cli/v10/configuring-npm/package-json#files


styfle:
Upon closer examination, it seems like the these top level assets are used at runtime so I'll close this.

PR #535, opened 2023-11-06
feat: add `wl` alias for `wolfram`

stone-zeng:
This pr adds `wl` alias for `wolfram`.


stone-zeng:
`wl` is the abbrevation of Wolfram Language, see <https://reference.wolfram.com/language/ref/format/WL.html>.

PR #538, opened 2023-11-15
fix(registry): loadLanguage does not await loadGrammar

### LINKED ISSUE #438: Highlighter's `loadLanguage` is buggy
Hi folks,

I was digging into https://github.com/banga/git-split-diffs/issues/43 and traced it down to me switching to `highlighter.loadLanguage` to load languages on demand to decrease start-up cost. It looks like the behavior of this method is pretty buggy.

Minimal repro here:
```ts
import shiki from 'shiki';

    theme: 'nord',
    langs: [],
});

await highlighter.loadLanguage('md');

// Throws "Error: No grammar provided for <source.json>"
await highlighter.loadLanguage('json');
```

If you remove the `await highlighter.loadLanguage('md')` line above, this no longer throws.

I tried to dig into it and haven't figure out the source, but this block of code seemed suspicious to me: https://github.com/shikijs/shiki/blob/870acb063e1ca0c2fb8ec15dc0e3991684c33ae7/packages/shiki/src/registry.ts#L59-L65 is passing an async callback to `reduce`. Unless there's something really clever going on here, this seems incorrect. The `acc` will end up being a `Promise` object with some string properties set on to it. Anyway, fixing this (by replacing with a for loop) doesn't actually fix the issue.

gregermendle:
This fix ensures that `loadGrammar` is awaited and prevents `loadLanguage` from throwing in registry.ts.

- [x] Add a test if possible

- [x] This PR fixes #438

### DIFF
--- a/packages/shiki/src/__tests__/issues.test.ts
+++ b/packages/shiki/src/__tests__/issues.test.ts
@@ -34,3 +34,15 @@ test(`Don't preload any language if lang is set to an empty array (#326)`, async
   })
   expect(highlighter.getLoadedLanguages()).toHaveLength(0)
 })
+
+// https://github.com/shikijs/shiki/issues/438
+test(`Highlighter's loadLanguage is buggy (#438)`, async () => {
+  const highlighter = await getHighlighter({
+    theme: 'nord',
+    langs: []
+  })
+
+  await highlighter.loadLanguage('md')
+  await highlighter.loadLanguage('json')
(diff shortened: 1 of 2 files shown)


gregermendle:
Updated this so its easier to test out if the fix actually worked. I also updated the code to make it a bit cleaner and hopefully easier to read. Running test on the first commit should result in an error saying a certain grammar (random for me) doesn't exist. Running it with the last commit results in no error and a passing test.

PR #543, opened 2023-11-28
fix: vue template syntax bug #503

### LINKED ISSUE #503: Vue directives and interpolation highlighting does not work
Vue directives and interpolation highlighting are handled by injections in https://github.com/vuejs/language-tools/blob/master/packages/vscode-vue/syntaxes/vue-directives.json and https://github.com/vuejs/language-tools/blob/master/packages/vscode-vue/syntaxes/vue-interpolations.json.

Shiki does not have those injections, so highlighting does not work.

nova1751:
- [x] This PR fixes #503

This bug is caused by fact that the `vue.tmLanguage.json` in [`Volar` repo](https://github.com/vuejs/language-tools/tree/master/extensions/vscode/syntaxes) doesn't support Vue directives and interpolations,the Volar uses `Injections` instead.The PR use `vue-html` syntax to render the code in tag `template` to fix the bug.

- the original：
  Snipaste_2023-11-28_20-27-18
- use `vue-html`： 
  Snipaste_2023-11-28_20-29-05

### DIFF
--- a/packages/shiki/languages/vue.tmLanguage.json
+++ b/packages/shiki/languages/vue.tmLanguage.json
@@ -861,7 +861,7 @@
               "include": "#html-stuff"
             }
           ]
-        }
+                }
       ]
     },
     "template-tag-2": {
@@ -911,10 +911,7 @@
           "include": "#template-tag"
         },
         {
-          "include": "text.html.derivative"
-        },
-        {
-          "include": "text.html.basic"
+          "include": "text.html.vue-html"
         }
       ]
     },


antfu:
The grammar is automatically fetched from upstream, so changing them will not fix it. Instead it's that Shiki is currently lacking for the injection support, which I fixed implemented it in Shikiji: https://github.com/antfu/shikiji/pull/48

nova1751:
> The grammar is automatically fetched from upstream, so changing them will not fix it. Instead it's that Shiki is currently lacking for the injection support, which I fixed implemented it in Shikiji: [antfu/shikiji#48](https://github.com/antfu/shikiji/pull/48)

Well,I get it.I fixed this thing by replacing the normal `html`'s rendering logic with the `vue-html`'s which could make elements render correctly in tag `template`.It works without injections actually,though it might be a bit weird to change the files fetched automatically.

PR #412, opened 2023-01-13
feat: add dynamic node_module resolution for wasm path

muenzpraeger:
- [ ] Add a test if possible
- [x] Format all commit messages with [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/)

- [x] This PR fixes #398

PR #557, opened 2024-01-26
feat!: merge Shikiji back into Shiki for v1.0

### LINKED ISSUE #33: Question: Recommendations on how to handle dark mode

### LINKED ISSUE #131: Configure package exports for browser build
I've been trying out the browser build @antfu added recently.

There is a slight issue in that we only specify the following in `package.json`:

```json
  "main": "dist/index.js",
  "module": "dist/index.esm.js",
```

which isn't necessarily true. `index.esm.js` is valid ESM **within nodejs**. there is a separate `index.browser.mjs` which is valid ESM for browsers.

however, rollup and most other tools will resolve to the node one because of the `module` field in the package manifest.

im not entirely sure how we resolve this as thats the correct behaviour for node, but not for browsers.

its true we could just tell browser consumers to directly import `index.browser.mjs` but we then need to alter the typescript declarations somehow to strongly type imports from that path...

```ts
import {Highlighter} from 'shiki'; // works
import {Highlighter} from 'shiki/dist/index.browser.mjs'; // doesn't work, TS has no clue what this file is as there's no index.browser.d.ts
```

### LINKED ISSUE #247: CloudFlare workers support
Hello @octref 😄

Thanks for this amazing project!

We are trying to use it as a syntax highlighter on the Docus project at NuxtLabs, which will most likely be ported as [@nuxt/content](https://content.nuxtjs.org/) v2 at some point.

We are using latest updates from Nuxt (Nuxt 3 beta and Nuxt Bridge) in that project, as this is made to be deployed on a CloudFlare worker environment.

While using Shiki, we encountered an issue in which it fallbacks to Node loader as it does not detect CloudFlare environment.

Here is the line that seem to be in cause:
https://github.com/shikijs/shiki/blob/main/packages/shiki/src/loader.ts#L49

Is CloudFlare Workers support planned for Shiki?

If so, maybe we could help on a PR supporting that use case!

cc: @farnabaz

### LINKED ISSUE #297: It there a way to support both diff and syntax highlighting at the same time?
For example I can use `diff` as language specifier to get a diff highlighting:

image

to get this:

```diff
+ var a = 1
- var 2 = 2
```

But I lose syntax highlight for other code.

```diff
+ var a = 1
- var 2 = 2

a = 3

function b() {}
```

or use `js` as language to get syntax highlight but you can't show diff highlight then:

```js
+ var a = 1
- var 2 = 2

a = 3

function b() {}
```

wondering is there a way to get both of them work just like the pull request diff view on Github?

image

### LINKED ISSUE #350: Suggestion: alternate syntax for language loading
👋 Hello! I’m a huge fan of this project and I’m excited for the direction this is taking. I’ve implemented Shiki in a number of different setups and wanted to suggest a possibly alternate approach to `setCDN()` as that’s been my biggest sticking point. Please close this issue if this suggestion is out-of-place or not the direction you’d want to take.

**Problem**

Shiki implementing its own `fetch()` mechanism and requiring a specific folder structure only works for specific environments, but makes it cumbersome when dealing with SSR, CloudFlare workers, or other setups. See the number of open issues filed around this:

- https://github.com/shikijs/shiki/issues/317
- https://github.com/shikijs/shiki/issues/247
- https://github.com/shikijs/shiki/issues/242
- https://github.com/shikijs/shiki/issues/216

**Proposal**

Now, the solution to all of these may just be “we’re holding it wrong” and additional docs or clarification are needed. But just for discussion, what if the following mechanism were used instead to load languages:

```diff
  shiki.getHighlighter({
    theme: 'nord',
    langs: [
      {
        id: 'rockstar',
        scopeName: 'source.rockstar',
-       path: './rockstar.tmLanguage.json' // or `plist`
+       import: () => import('./rockstar.tmLanguage.json')
      }
    ]
  })
```

It’s a tiny change, but it would mean that no internal `fetch()` mechanism would be needed, nor would that need to be re-implemented for SSR, nor mocked in testing, etc. etc. Now that `import()` is a part of the language spec, it passes off responsibility to the platform to handle loading that module, and it can make better optimizations around this than shiki can (also, if a person still wanted to continue using the CDN structure in a `/public/` directory this would let them use it the same way—both local and remote URLs are supported by `import()`).

Of course, the outlier is the oniguruma Wasm file. But I believe the `fetch()` still wouldn’t be needed as only a URL is required, and JS itself can handle the rest:

```diff
  shiki.getHighlighter({
    theme: 'nord',
    wasmURL: import.meta.env.WASM_URL, // "/assets/oniguruma.wasm"
  })
```

This would also let the build tool better optimize this, and even pass environment-dependent values for this Wasm URL (and I think by this point many people understand that an absolute URL is required, and also build tools like Vite allow management of this through [`?url` and `?worker` imports](https://vitejs.dev/guide/assets.html#importing-asset-as-url)).

**Summary**

- `path` string replaced with async `import()` (wrapped in a func so Shiki controls when this fires)
- No internal `fetch()`
- Expose Wasm URL in options

My belief is that this would simplify Shiki and make it more flexible to use on more platforms. But is that the case? Am I missing something? Would love some holes poked in this theory. Thanks for reading! 🙂

### LINKED ISSUE #357: ESM Support for Shiki (usage in Framer)
Hello! I’d love to support Shiki within Framer, as part of a component such as https://syntax.page/

Unfortunately, it appears Shiki isn’t ESM-compatible. In Framer, I can pull in NPM packages to create custom React components with, but only if they are ESM-compatible. https://www.framer.com/npm/

This is what it errors on:

image

`T10:04:20.356Z warn: [modules-runtime] Error in 'codeFile/Shiki.tsx': SyntaxError: The requested module 'https://ga.jspm.io/npm:vscode-oniguruma@1.6.2/release/main.js' does not provide an export named 'createOnigScanner'
(anonymous) @ instrument.js:91
instrument.js:91 `

Would appreciate any guidance or tips.

### LINKED ISSUE #380: Add plugin capability
Shiki would benefit from providing a pluggable plugin architecture.

*Why?*
- Modifications to the current HTML structure are only possible if someone uses `renderHTML` directly and provides their element definition.
- Modifications can be shared, but there are not built-in merge mechanisms if someone wants to apply their own modifications.
- By having plugins it frees project owners/contributors from work/requests - others can built themselves. ;-)
- It builds an ecosystem.

*How?*
Good question. I started to design an (opinionated) approach, you can look at it in #381

### LINKED ISSUE #411: Catppuccin themes
Would you be open to a PR for the [catppuccin themes](https://github.com/catppuccin)?

I wanted to ask before opening, as it seems like new themes are [no longer being encouraged](https://github.com/shikijs/shiki/blob/main/docs/themes.md#not-actively-adding-new-themes)?

However, I was hoping this may fall under the "rarely accept" category. 🙂

### LINKED ISSUE #424: Plan for 1.0
Well, I haven't been spending much time on OSS for a while, and now I plan to get back to it. So here's a plan for 1.0. I'm guessing it'll take 2 months. This is a brain dump list and will change as things goes.

- Website: redo it. We have too many content there and it's hard to add new ones.
- Docs: redo it. Give it a better organization and make it easy to add new things.
- API: review the use cases and simplify the surface.
- New features
  - Plugins: do it as part of 1.0. #381
  - ESM support. 
  - Dark theme (multiple theme) support
  - Line number support #3 
- Examples: make one for each major framework and keep it in the core repo.
- Samples: make sure each language has one.
- Theme/Grammar autofetching: auto-generate a changelog. Auto-open issue for added/deleted theme/grammar
- Infrastructure: 
  - publish a nightly build #57
  - automate changelog generation #223

### Todo – Meta

- [x] Label all issues. Close ones without repro.
- [x] Merge bug-fix PRs.
- [x] Fix valid bugs.
- [x] Release one last 0.x version. Feature freeze.
- [ ] Do the overhauling in `dev` branch
- [ ] Merge `dev` back to main for releasing 1.0

### Todo – Open to contributions

- [ ] Examples: Create a vanilla/nextjs/vite official example
  - [x] Next 13: https://github.com/shikijs/next-shiki
  - [ ] Vite
  - [ ] Vanilla

### LINKED ISSUE #425: Terminal Renderer
This should render `themedToken[][]` to a string that contains ANSI codes.

Additionally, with core support for `diff`, we should be able to generate syntax-highlighted diffs. Maybe one day we could be a git pager!

### LINKED ISSUE #438: Highlighter's `loadLanguage` is buggy

### LINKED ISSUE #473: Remove the default loading of `nord` theme when `getHighlighter`

### LINKED ISSUE #491: Multiple grammars for the same lanuage
Hi,

The nextflow language comes with several cross-referencial grammars: https://github.com/nextflow-io/vscode-language-nextflow/tree/master/syntaxes. 

How can I add them to shiki, so they are included correctly?

### LINKED ISSUE #509: Tokens includeExplanation typings mismatch
## Description

While trying to optimize performance of great shiki we encountered confusing typings.
`ThemedTokenizer` and [types](https://github.com/shikijs/shiki/blob/e4d08552d20fa808a28ff9967f0604c934c8060e/packages/shiki/src/types.ts#L271) both states default of `false` for `includeExplanation` options.
But [highlighter.codeToThemeTokens](https://github.com/shikijs/shiki/blob/e4d08552d20fa808a28ff9967f0604c934c8060e/packages/shiki/src/highlighter.ts#L150) is overriding `includeExplanation` default to `true`, while being typed with mentioned default `false` thus being wrongly docummented while accessed in code.

## Expected behaviour
Either codeToThemeTokens having `includeExplanation: false` default (would make sense,  its performance save was mindblowing), or fixing types to have correct typing (different defaults) for both `ThemedTokenizer` and `highlighter.codeToThemeTokens` (eg. for compatibility reasons I can see).

Guess should be quick change, but does not feel aligned with project enough to pick solution myself and create PR

### LINKED ISSUE #510: CDN/Assets-free ESM Approach of loading wasm and languages
Shiki's assets fetching part is always a bit tricky to set up, we introduced `setCDN`, environment detection, trying to make it work in various environments. At the same time, users who use bundlers might still need to manually handle the asset distribution and the potential misalignment in dev/build environment.

Those are tricky but still solvable if you take some time. However, we hit the wall when trying to deploy shiki to Cloudflare Workers. In CF workers, or other worker runtimes, you usually don't have access to Node.js's API, meaning no file system, etc. And loading WASM can also be a bit challenging.

This is mainly due to the fact that we are resolving the assets at runtime, meaning the bundlers won't know the relationship between your code and assets, making the connection easy to break when moving things around.

Thus, I made an explorational rewrite as https://github.com/antfu/shikiji, trying to improve the situation.

Basically, what it does is to convert the JSON languages and themes into ESM module `.mjs` and then reference them using dynamic `import()` ([code here](https://github.com/antfu/shikiji/blob/main/src/assets/langs.ts)). And inline the WASM into Base64 so it can be an ESM module as well. With these, we are building the relationship in pure ESM without any assets and do not require any Node API or network requests. Not only would this make the codebase simpler and the data loading more efficient, but it also would allow bundlers to generate corresponding chunks and ship the data automatically.

The result turns working very well for us, and we are able to run `shikiji` on Cloudflare worker as an on-demand syntax highlighter running on edge.

Since we are in the process of preparing 1.0, I wonder if you think it would make sense to have this approach as a part of shiki core? /cc @octref @orta

### LINKED ISSUE #542: Potential vscode-oniguruma issue when running shiki on client-side with docusaurus 3.0 and webpack 5
We are a open-source organisation and we are migrating our documentation to docusaurus 3.0. We only have TextMate grammar definition and Shiki appears to be the best choice for syntax highlighting.

I am aware of the existence of https://github.com/shikijs/twoslash, but it runs on server side and I am wondering if it is possible to have something that runs on the client side.

I have written a very small client-side wrapper like this (the rest is docusaurus 3.0 from docuaurus.new which I run on GH codespaces; this file is under `src/components/Shiki/index.ts):
```ts
import * as shiki from 'shiki';

import react from 'react';

export const ShikiHighlight = ({ lang, rawcode, props }: {
  lang: string;
  rawcode: string;
  props: unknown;
}) => {
  const [code, setCode] = react.useState<string>('');

  react.useEffect(() => {
    (async () => {
      try {
        // const onig = await fetch("https://cdn.jsdelivr.net/npm/vscode-oniguruma@2.0.1/release/onig.wasm");
        // shiki.setWasm(onig);
        const highlighter = await shiki.getHighlighter({ theme: 'nord' });
        setCode(highlighter.codeToHtml(rawcode, { lang }));
      } catch (e) {
        console.log(e);
      }
    })();
  }, [lang, rawcode]);

  return <div dangerouslySetInnerHTML={{ __html: code }} />;
};
```

Upon using this component, it shows 
`WebAssembly: Response has unsupported MIME type 'text/html; charset=utf-8' expected 'application/wasm'` on Firefox, or `Failed to execute 'compile' on 'WebAssembly': Incorrect response MIME type. Expected 'application/wasm'.` on chromium.

Copying `onig.wasm` to `/static` wouldn't fix this issue.

If I fetch and specify `vscode-oniguruma` wasm instead (that is, removing the commented lines), shiki will still error:
`import object field '_embind_register_constant' is not a Function` on Firefox, or `LinkError: WebAssembly.instantiate(): Import #3 module="env" function="_embind_register_constant": function import requires a callable` on Chromium.

Finally, if I use `vscode-oniguruma@1.6.2` as instructed here: https://github.com/shikijs/shiki/issues/382#issuecomment-1340987499, it will instead show `LinkError: WebAssembly.instantiate(): Import #4 module="env" function="setTempRet0": function import requires a callable`.

I am puzzled. Is this the expected behaviour when trying to run shiki on browser with oniguruma wasm, or is there anything related to shiki which I did wrong? I hope that I asked this question at the correct place, and I apologize if I didn't.

### LINKED ISSUE #551: Move grammars/themes update script as a seperate package
Currently, we serve all those grammar and themes update logics inside this repo and create quite some noise to the commit history. And since those themes can also be reused by other integrations like Monaco editors, it would be nice to extra those logics as separate packages so they can be better reused and well organized.

I actually already did the work by porting some scripts from Shiki:
- https://github.com/antfu/textmate-grammars-themes/tree/main/packages/tm-grammars
- https://github.com/antfu/textmate-grammars-themes/tree/main/packages/tm-themes

The rewrite would also help us to obtain more info and properly credit the original authors with their [licenses](https://github.com/antfu/textmate-grammars-themes/blob/main/packages/tm-grammars/NOTICE).

Meanwhile, the packages are automatically upgraded and released. So if we end up migrating to using it as the source, we only need to bumper the dependencies periodically to get the latest grammar/themes.

I would prepare the PR if the team agreed on that. I am also open to transferring that repo into this org.

Related:
- https://github.com/shikijs/shiki/issues/57#issuecomment-1407411941

### LINKED ISSUE #553: Global fetch types overriding browsers DOM apis
The https://github.com/shikijs/shiki/pull/504 moved the declare for `globalThis.fetch` to publish it with the lib, however, this means that any instance of me using `fetch(result => result.text())` resolves to `Promise<any>` whereas before it would resolve to `Promise<string>` which is obviously much more preferred. 

The types for node have been including fetch results (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/66824) which would be a better solution instead of polluting the global type space.

### LINKED ISSUE #555: Recent release missing some language changes
I recently installed 0.14.7 and I noticed a change https://github.com/shikijs/shiki/pull/446 wasn't reflected in `import { BUNDLED_LANGUAGES } from "shiki";`

If you do a search like:

```tsx
const selectedLang = BUNDLED_LANGUAGES.find(
      (l) => l.aliases?.includes("protobuf")
);
```

It comes up empty despite that PR adding the language alias. I don't know if this was caused by a faulty cache while publishing or what.

### LINKED ISSUE #556: various files are not found when used in Sveltekit
I am trying to do some syntax highlighting on markdown blog site. I used marked-highlight in conjunction with shiki to achieve this in sveltekit.

```ts
import {Marked} from 'marked'
import {getHighlighter} from 'shiki'

export const markedShiki = new Marked(
	markedHighlight({
        async: true,
		async highlight(code, lang) {
			const highlighter = await getHighlighter({
				theme: 'nord'
			});
			const html = highlighter.codeToHtml(code, { lang });
			return html;
		}
	})
);

//+page.ts

async function load({params}){
  ...
  return await marked.parse(txt)
}
```

This sometimes will load the code perfectly for 1 second before telling me that it has failed. With the error looking like this:

```
SvelteKitError: Not found: /themes/nord.json
    at resolve (C:\Users\leung\OneDrive\Desktop\Programming\mdsvex\markdown-blog-mdsvex-ts\node_modules\@sveltejs\kit\src\runtime\server\respond.js:495:13)
    at resolve (C:\Users\leung\OneDrive\Desktop\Programming\mdsvex\markdown-blog-mdsvex-ts\node_modules\@sveltejs\kit\src\runtime\server\respond.js:295:5)
    at #options.hooks.handle (C:\Users\leung\OneDrive\Desktop\Programming\mdsvex\markdown-blog-mdsvex-ts\node_modules\@sveltejs\kit\src\runtime\server\index.js:63:56)
    at Module.respond (C:\Users\leung\OneDrive\Desktop\Programming\mdsvex\markdown-blog-mdsvex-ts\node_modules\@sveltejs\kit\src\runtime\server\respond.js:292:40)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  status: 404,
  text: 'Not Found'
}
SvelteKitError: Not found: /dist/onig.wasm
    at resolve (C:\Users\leung\OneDrive\Desktop\Programming\mdsvex\markdown-blog-mdsvex-ts\node_modules\@sveltejs\kit\src\runtime\server\respond.js:495:13)
    at resolve (C:\Users\leung\OneDrive\Desktop\Programming\mdsvex\markdown-blog-mdsvex-ts\node_modules\@sveltejs\kit\src\runtime\server\respond.js:295:5)
    at #options.hooks.handle (C:\Users\leung\OneDrive\Desktop\Programming\mdsvex\markdown-blog-mdsvex-ts\node_modules\@sveltejs\kit\src\runtime\server\index.js:63:56)
    at Module.respond (C:\Users\leung\OneDrive\Desktop\Programming\mdsvex\markdown-blog-mdsvex-ts\node_modules\@sveltejs\kit\src\runtime\server\respond.js:292:40)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  status: 404,
  text: 'Not Found'
}
SvelteKitError: Not found: /themes/nord.json
    at resolve (C:\Users\leung\OneDrive\Desktop\Programming\mdsvex\markdown-blog-mdsvex-ts\node_modules\@sveltejs\kit\src\runtime\server\respond.js:495:13)
    at resolve (C:\Users\leung\OneDrive\Desktop\Programming\mdsvex\markdown-blog-mdsvex-ts\node_modules\@sveltejs\kit\src\runruntime\server\respond.js:295:5)
    at #options.hooks.handle (C:\Users\leung\OneDrive\Desktop\Programming\mdsvex\markdown-blog-mdsvex-ts\node_modules\@sveltejs\kit\src\runtime\server\index.js:63:56)
    at Module.respond (C:\Users\leung\OneDrive\Desktop\Programming\mdsvex\markdown-blog-mdsvex-ts\node_modules\@sveltejs\kit\src\runtime\server\respond.js:292:40)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  status: 404,
  text: 'Not Found'
}

```

The list is a lot longer but it seems to not be able to find the files. I made sure that they are definitely in node modules.

antfu:
This PR merges https://github.com/antfu/shikiji into the repo, renamed back to Shiki. All other sub-packages are now placed under `@shikijs` npm scope.

## Notable Changes:

- Core changes are listed https://shikiji.netlify.app/guide/compat#breaking-changes-from-shiki, will update the page and provide more detailed migration path in the new docs.
- VuePress plugin is removed, as VuePress itself is discontinued. Its successor, VitePress, uses the new implementation by default.
- `shiki-renderer-path` and `shiki-renderer-svg` are removed due to the low usage. We are open to bringing it back if there is demand.
- Themes and grammar files are now live in a separate repository: https://github.com/shikijs/textmate-grammars-themes#add-a-new-grammar
- The old site gets removed in favor of the new docs

### Notable

- resolve #424
- resolve #510
- resolve #33
- resolve #381
- resolve #412
- resolve #551
- resolve #247
- resolve #350
- resolve #380
- resolve #131
- resolve #425

### Bugs

- fix #384
- fix #438
- fix #556
- fix #543
- fix #553
- fix #555
- fix #542
- fix #509
- fix #491
- fix #411 
- fix #357
- fix #297
- fix #473

## TODO

- [x] Revise overall documentation and phrasing
- [x] Improve contribute guide
- [x] Better migration guide


antfu:
Merge it for now, and let's iterate on main! GitHub is not very performant when dealing with grant PR like this.

DominusKelvin:
Great job @antfu 👏🏾

PR #571, opened 2024-01-31
feat: add aliases for fortran formats

uncenter:
### Description

Adds aliases for the two forms of Fortran.

### Linked Issues

N/A

### Additional context

N/A


antfu:
This file is generated. Alias is stored in tm-grammars, please send PR there instead. Thanks!

PR #585, opened 2024-02-11
fix: remove trailing newline added by markdown-it

KermanX:
### Description

**Scope: `@shikijs/markdown-it`**

Shiki transpiles blank lines into `<span class="line"></span>`. And the source code received in `markdownit.options.highlight` has a trailing new line. This results in a blank line in the rendered HTML.

This PR fixes this by removing the last character of the source code.

**Example input**:
```markdown
<!-- Blank -->
'''ts
'''

<!-- One blank line -->
'''ts

'''

<!-- Normal code -->
'''ts
123
'''
```

**Output** (simplified):
```html
<!-- Blank -->
<pre><code>
  <span class="line"></span>
</code></pre>

<!-- One blank line -->
<pre><code>
  <span class="line"></span>
</code></pre>

<!-- Normal code -->
<pre><code>
  <span class="line">123</span>
</code></pre>
```

### Linked Issues

fix https://github.com/slidevjs/slidev/issues/1276


packages/markdown-it/src/core.ts
```diff
@@ -50,7 +50,7 @@ export function setupMarkdownIt(
     })
 
     return highlighter.codeToHtml(
-      code,
+      code.slice(0, -1),
```
antfu, line 53:
```suggestion
      code.trimEnd(),
```

KermanX: I was wondering maybe there will be someone who intended to have a blank line or whitespaces at the end of the code?🤔

antfu: I found that VitePress is doing this mandatory: https://github.com/vuejs/vitepress/blob/c582a8d5fd82b84d412c7e6c84e74faeb23beac6/src/node/markdown/plugins/highlight.ts#L144

I wonder why markdown-it is doing this (if it always adds a tailing newline) - do you have a reference? `.slice(0, -1)` looks a bit magic and unreliable to me as it might strip out real code if the tailing newline is not a determined behavior.

KermanX: I think markdown-it always adds a tailing newline, bacause:

https://github.com/markdown-it/markdown-it/blob/master/lib/rules_block/fence.mjs#L89:
```js
  token.content = state.getLines(startLine + 1, nextLine, len, true)
//                                                              ^ "keepLastLF"
```

and https://github.com/markdown-it/markdown-it/blob/master/lib/renderer.mjs#L43:
```js
    highlighted = options.highlight(token.content, langName, langAttrs) || escapeHtml(token.content)
```

PR #588, opened 2024-02-15
feat: add length and time limit to tokenizer

ije:
### Description

Fix the textmate tokenizer freeze/crash the browser if a line is too long:

- Set the `timeLimit` parameter of `tokenizeLine2` to 500ms (align to vscode)
- Do not attempt to tokenize if a line is too long (enable by adding the `maxTokenizationLineLength` opion)

### Linked Issues

_

### Additional context

_Refs_

`timeLimit`:
https://github.com/microsoft/vscode/blob/3ce7ccff4116da34370cecfb329420f266499ec4/src/vs/workbench/services/textMate/browser/tokenizationSupport/textMateTokenizationSupport.ts#L54

Do not attempt:
https://github.com/microsoft/vscode/blob/6722b81416d8f1ee24209ca4a19792b6e6a55beb/src/vs/workbench/services/textMate/browser/tokenizationSupport/tokenizationSupportWithLineLimit.ts#L38

### DIFF
--- a/packages/core/src/code-to-tokens-base.ts
+++ b/packages/core/src/code-to-tokens-base.ts
@@ -45,6 +45,11 @@ export function tokenizeWithTheme(
     ...options?.colorReplacements,
   }
 
+  const {
+    tokenizeMaxLineLength = 0,
+    tokenizeTimeLimit = 500,
+  } = options
+
   const lines = splitLines(code)
 
   let ruleStack = INITIAL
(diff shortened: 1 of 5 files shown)


packages/core/src/code-to-tokens-base.ts
```diff
@@ -59,6 +59,19 @@ export function tokenizeWithTheme(
       continue
     }
 
+    // Do not attempt to tokenize if a line is too long
+    const maxTokenizationLineLength = 20000
```
antfu, line 63: I am not sure about this. It makes sense to do it in Monaco, but for Shiki core, I think the users are responsible for controlling what passes in. We shouldn't disallow ppl to do so if they want deliberately.

ije: do you like an option? (default is disable)

antfu: OK then 👍

ije: done, please check the testing.

PR #605, opened 2024-02-29
fix: add `jl` alias for Julia

Odilf:
### Description

Julia files have a `.jl` extension, so it would be a good idea to add a `jl` alias for the language.

This is useful for me because I'm working on a project where I want to highlight dynamically based on the file extension, and I suspect this is hardly uncommon. 

This could be done manually today by consumers of the library with `langAlias` but I think it's reasonable to include the main file extension for Julia. 

### Linked Issues

N/A

### Additional context

No tests are provided because as far as I can tell languages aliases aren't tested. Besides, I think it's a very simple change so it's easy to tell if it's correct or not. 

Thanks!


Odilf:
Netlify deploy is failing due to typo in a link, seems to be due to what was fixed in 0ab9a2d

PR #612, opened 2024-03-05
docs: mention that the user is responsible for adding transformer classes css

francisko-rezende:
### Description

Explicitly mention that the user is in charge of adding the css rules for classes added by transformers in the relevant section of the docs.

### DIFF
--- a/docs/packages/transformers.md
+++ b/docs/packages/transformers.md
@@ -36,6 +36,8 @@ const html = await codeToHtml(code, {
 
 ## Transformers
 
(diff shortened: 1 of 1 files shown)


docs/packages/transformers.md
```diff
@@ -36,6 +36,8 @@ const html = await codeToHtml(code, {
 
 ## Transformers
 
+Transformers add specific CSS classes to relevant tags but do not come with any styles, the users is responsible for writing the CSS rules for the classes themselves.
```
antfu, line 39:
```suggestion
Transformers add specific CSS classes to relevant tags but do not come with styles, you might want to provide your own CSS rules to style them properly.
```

PR #645, opened 2024-03-29
perf(shiki-monaco): improve tokenizer performance

ije:
### Description

current implement the `tokenize` method computes `colorMap` and `colorToScopeMap` every time it's called.
the patch moved the `colorMap` out of the `tokenize` method, only re-computes when `setTheme` is called.

no testing, but the code works in my wip editor:

https://github.com/shikijs/shiki/assets/2883484/28c99b23-ffff-4ed1-b27d-c41d41e56db8

### DIFF
--- a/packages/monaco/src/index.ts
+++ b/packages/monaco/src/index.ts
@@ -1,16 +1,9 @@
-import type * as monaco from 'monaco-editor-core'
+import type monacoNs from 'monaco-editor-core'
 import type { ShikiInternal, ThemeRegistrationResolved } from '@shikijs/core'
 import type { StateStack } from '@shikijs/core/textmate'
 import { INITIAL, StackElementMetadata } from '@shikijs/core/textmate'
 
-type Monaco = typeof monaco
-
-export interface MonacoInterface {
-  editor: Monaco['editor']
-  languages: Monaco['languages']
-}
-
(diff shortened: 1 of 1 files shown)


packages/monaco/src/index.ts
```diff
@@ -118,38 +117,32 @@ export function shikiToMonaco(
...
       })
     }
   }
 }
 
-class TokenizerState implements monaco.languages.IState {
+class TokenizerState implements monacoNs.languages.IState {
   constructor(
     private _ruleStack: StateStack,
-    public highlighter: ShikiInternal<any, any>,
```
ije, line 140: we don't need the `highlighter` field, just use the colsure one

packages/monaco/src/index.ts
```diff
@@ -166,9 +159,12 @@ function normalizeColor(color: string | undefined): string | undefined
 function normalizeColor(color: string | undefined) {
   if (!color)
     return color
-  color = color.replace('#', '').toLowerCase()
+
+  color = (color.charCodeAt(0) === 35 ? color.slice(1) : color).toLowerCase()
```
ije, line 163: this should be faster by checking first char code instead of calling replace

antfu (approved):

Issue #691, opened 2024-06-02
Rehype: default language


fuma-nama:
### Clear and concise description of the problem

When no language specified, the codeblock will not be processed by Shiki.

### Suggested solution

Add a `defaultLanguage` option to Rehype Integration, when no language specified, use the default language as a fallback.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/eslint-stylistic/eslint-stylistic/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://eslint.style/contribute/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Contributes

- [X] If this feature request is accepted, I am willing to submit a PR to fix this issue

PR #692, opened 2024-06-02
feat(rehype): Support default language option

fuma-nama:
### Description

As the `markdown-it` integration also provides similar options, this PR adds the `defaultLanguage` option to the rehype integration.

When no language specified, it uses the default language instead.
Note that it doesn't change the current behaviour. When neither language nor default language is provided, the code block will not be processed by Shiki, hence you can handle it on your own, and won't break existing usages.

### Linked Issues

#691 

### Additional context

unit tests are added.

### DIFF
--- a/packages/rehype/src/core.ts
+++ b/packages/rehype/src/core.ts
@@ -1,7 +1,7 @@
 import type { CodeOptionsMeta, CodeOptionsThemes, CodeToHastOptions, HighlighterGeneric, TransformerOptions } from 'shiki/core'
 import type { Element, Root } from 'hast'
 import type { BuiltinTheme } from 'shiki'
-import type { Plugin } from 'unified'
+import type { Transformer } from 'unified'
 import { toString } from 'hast-util-to-string'
 import { visit } from 'unist-util-visit'
 
@@ -17,6 +17,11 @@ export interface RehypeShikiExtraOptions {
    * @default false
    */
   addLanguageClass?: boolean
+
+  /**
+   * The default language to use when is not specified
+   */
+  defaultLanguage?: string
 
   /**
    * Custom meta string parser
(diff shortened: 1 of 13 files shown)


antfu:
Could you also help to implement the same interface as in https://github.com/shikijs/shiki/pull/689? Thanks!

fuma-nama:
Added `fallbackLanguage` option

packages/rehype/src/index.ts
```diff
@@ -17,24 +17,22 @@ export type RehypeShikiOptions = RehypeShikiCoreOptions
   }
 
 const rehypeShiki: Plugin<[RehypeShikiOptions], Root> = function (
-  options = {} as any,
+  options = { theme: 'vitesse-dark' },
```
antfu, line 20: I don't think we should provide a default value for this

packages/rehype/src/core.ts
```diff
@@ -57,18 +67,21 @@ declare module 'hast' {
...
     parseMetaString,
     cache,
+    defaultLanguage,
+    fallbackLanguage,
+    onError,
     ...rest
-  } = options
-
-  const prefix = 'language-'
+  }: RehypeShikiCoreOptions,
```
antfu, line 82: Emm, why this refactoring? I prefer not to destructure the options inside the argument.

packages/rehype/src/core.ts
```diff
@@ -86,20 +99,15 @@ const rehypeShikiFromHighlighter: Plugin<[HighlighterGeneric<any, any>, RehypeSh
...
 
-      const classes = head.properties.className
-
-      if (!Array.isArray(classes))
-        return
-
-      const language = classes.find(
-        d => typeof d === 'string' && d.startsWith(prefix),
-      )
+      let lang = getLanguage(head) ?? defaultLanguage
```
antfu, line 102: If this function only gets used once, I'd think it's better not to extract it.

packages/rehype/src/index.ts
fuma-nama: But casting it to `any` type may be unnecessary here. Perhaps we can use `as RehypeShikiOptions` instead

antfu (approved):

PR #695, opened 2024-06-07
feat(twoslash): add line query rendering option for twoslash renderer

### LINKED ISSUE #688: Twoslash node query blocks (^?) no longer display on separate lines since 0.x
### Validations

### Describe the bug

In 0.*, twoslash node query blocks (^?) were shown on a separate line, like here in the TypeScript docs.

image

In 1.*, they are shown with position: absolute over the top:

image

This appears to be a [deliberate change](https://x.com/antfu7/status/1796577913963503808) to the renderer.

I can see the use cases for this new absolute positioning. But for my use cases (screenshotting and docs) I need the query blocks to be shown on a separate line to the nodes they target. A config option to go back to the old behavior would be appreciated.

I can see that rendererClassic is available for me, so I'm using that now - but a separate option to preserve the separate-line queries would be good.

### Reproduction

No reproduction required, since Anthony described this is a deliberate change

### Contributes

hddhyq:
### Description

add line query rendering option for twoslash renderer

### Linked Issues

fix #688


antfu:
Great, thank you!

PR #712, opened 2024-06-27
feat: introduce `GrammarState`

antfu:
### Motivations 

Sometimes, we might want to highlight a partial of the code snippet, which might not necessarily be a full language statement. 

For example, if we want to highlight a TypeScript type annotation `Pick<MyObject, string>[]` we might do:

```ts
import { codeToHtml } from 'shiki'

const highlighted = await codeToHtml('Pick<MyObject, string>[]', { lang: 'ts', theme: 'github-dark' })
```

We will get the result like this:

image

This isn't accurate. This is because from the grammar, Shiki doesn't know it's a type annotation already.

Instead, we could append `let a: ` to our code to hint that:

```ts
import { codeToHtml } from 'shiki'

const highlighted = await codeToHtml('let a: Pick<MyObject, string>[]', { lang: 'ts', theme: 'github-dark' })
```

Which will now render:

Screenshot 2024-06-27 at 18 25 47

While the highlighting is now correct, we have another challenge: removing the "temporary snippet" we just added from the highlighted HTML.

### Solutions

This PR introduces a new concept, `GrammarState`, with a few APIs around it. `GrammarState` is a special token that holds the grammar context and allows you to highlight from an intermediate grammar state, making it easier to highlight code snippets.

For example, you can get the grammar state with the `getLastGrammarState` method, and pass into the `grammarState` option:

```ts
import { createHighlighter } from 'shiki'

const shiki = await createHighlighter({ langs: ['ts'], themes: ['github-dark'] })

const stateTypeAnnotation = shiki.getLastGrammarState('let a:', { lang: 'ts', theme: 'github-dark' })

const highlightedType = shiki.codeToHtml(
  'Pick<MyObject, string>[]',
  {
     lang: 'ts', 
     theme: 'github-dark',
     grammarState: stateTypeAnnotation // <--- this
  }
)
```

Now Shiki would highlight correctly as it knows to start as the type annotation. You can keep that grammar state object for multiple uses as well.

image

---

For one-off grammar context shifting, we also provide a shorthand by the `grammarContextCode` option:

```ts
const highlightedType = shiki.codeToHtml(
  'Pick<MyObject, string>[]',
  {
     lang: 'ts', 
     theme: 'github-dark',
     grammarContextCode: 'let a:' // same as above, a temporary grammar state is created internally
  }
)
```

### Credits

Thanks to @benjamincanac for the idea and discussions

### DIFF
--- a/packages/core/src/code-to-tokens-base.ts
+++ b/packages/core/src/code-to-tokens-base.ts
@@ -1,12 +1,14 @@
 /* ---------------------------------------------------------
  * Copyright (C) Microsoft Corporation. All rights reserved.
  *-------------------------------------------------------- */
-import type { IGrammar, IRawThemeSetting } from './textmate'
+import type { IGrammar, IRawThemeSetting, StateStack } from './textmate'
 import { INITIAL } from './textmate'
 import type { CodeToTokensBaseOptions, FontStyle, ShikiInternal, ThemeRegistrationResolved, ThemedToken, ThemedTokenScopeExplanation, TokenizeWithThemeOptions } from './types'
 import { StackElementMetadata } from './stack-element-metadata'
 import { applyColorReplacements, isNoneTheme, isPlainLang, resolveColorReplacements, splitLines } from './utils'
 import { tokenizeAnsiWithTheme } from './code-to-tokens-ansi'
+import { ShikiError } from './error'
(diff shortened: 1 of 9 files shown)


orta:
Cool idea

PR #717, opened 2024-07-16
docs: add eleventy integration tutorial

alexandrehtrb:
Hello,

Thanks for this project!

This PR is for docs on Shiki integration with Eleventy (11ty).

I tried to run the docs locally, but it didn't work, it opened an URL for [undefined npm package](https://www.npmjs.com/package/undefined). I executed:

```sh
pnpm i
pnpm build
pnpm docs
```

Running on Windows.


antfu:
Thank you for your effort! But the point is that we could not document usage for **every single tool** - meaning that we only document usage of official packages - not usages. I would suggest you post this as your blog post, or somewhere with a shareable link - and then we are happy to link to it as an external resource.

Thanks

PR #718, opened 2024-07-20
fix: codeToTokens's includeExplanation option

uxiew:
## Description

The options's `includeExplanation` field in `codeToTokens` doesn't work!


packages/core/src/code-to-tokens-themes.ts
```diff
@@ -34,6 +34,10 @@ export function codeToTokensWithThemes(
           offset: _token.offset,
         }
 
+        if ('includeExplanation' in options) {
```
antfu, line 37:
```suggestion
        if ('includeExplanation' in options && options.includeExplanation) {
```

PR #721, opened 2024-07-24
chore: update doc about @shikijs plugins

fwqaaq:
### Description

Add @shikijs plugins CDN link.

I don't know if Node supports CDN, and haven't added markdown-it?

### DIFF
--- a/docs/packages/monaco.md
+++ b/docs/packages/monaco.md
@@ -10,6 +10,8 @@ Heavily inspired by [`monaco-editor-textmate`](https://github.com/zikaari/monaco
 
 ## Install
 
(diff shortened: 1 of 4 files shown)


antfu:
Thank you. While I like esm.sh a lot, it's not the only CDN provider out there, and honestly, it isn't our responsibility to document it on every package - it should be part of the URL convensions.

fwqaaq:
> Thank you. While I like esm.sh a lot, it's not the only CDN provider out there, and honestly, it isn't our responsibility to document it on every package - it should be part of the URL convensions.

If you publish it on `jsr`, you may not need to consider it. But when I use it in `Deno`, I would perfer to be able to directly get  CDN URL, rather than having to search for each one individually.

Issue #589, opened 2024-02-15
`Notation` transfomers removes empty annotated lines


lachieh:
### Validations

- [X] Follow our [Code of Conduct](https://github.com/shikijs/shiki/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://github.com/shikijs/shiki/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

### Describe the bug

It seems that all of the notation transformers like `transformerNotationHighlight` and `transformerNotationDiff` that if the line is empty, the `span.line` is removed entirely.

For example:
```js
hello.world() // [!code highlight]
// [!code highlight]
```
Results in:
```html
<pre>
  <code>
    <span class="line highlighted"><span>hello</span><span>.</span><span>world</span><span style="color:#666666">()</span></span>

  </code>
</pre>
```

But I would assume it should be:
```html
<pre>
  <code>
    <span class="line highlighted"><span>hello</span><span>.</span><span>world</span><span style="color:#666666">()</span></span>
    <span class="line highlighted"></span>
  </code>
</pre>
```

### Reproduction

https://stackblitz.com/edit/vitejs-vite-axpred?file=main.js

### Contributes

- [ ] I am willing to submit a PR to fix this issue
- [ ] I am willing to submit a PR with failing tests

antfu:
This is intentional and expected behavior. I am ok with exposing an option to toggle that, pr welcome

lachieh:
Thanks for the reply. It feels a little unexpected since the `span.line` is not removed unless there is a notation comment. Is this documented?

How would you indicate a diff that includes a new line as part of the code change?

lachieh:
Thanks, that gets me across the line for now.

I still think that this is probably still a bug since it is doing the opposite of what the notation comment is suggesting it will. Combined with the `transformerRemoveLineBreak` transformer and a `.line { display: block }` then the output wraps onto the previous line.

I'd like to suggest that an option to toggle is added, but the default behavior be that it just does what the comment says it will. I'm happy to submit a PR.

lachieh:
Small follow-up to document behavior. The multi-line notation `// [!code highlight:2]` also removes the starting `span.line` if the comment is on an otherwise empty line.

rishi-raj-jain:
Hey, I am writing this code block with Shiki:
[...]

PR #729, opened 2024-07-28
feat(rehype): trim trailing new line passed in from rehype/remark

Ovyerus:
### Description

Trim the single trailing newline at the end of code that gets passed into Shiki from Rehype, which would result in dangling empty `<span class="line">` elements at the end of *every* code block.

DOM screenshot, showing a completely empty `line` element after one expected line of code.
Markdown source screenshot, showing only one line of text.

Usually, these extra elements aren't an issue—evident by the fact that this appears to mostly have flown under the radar since the Rehype plugin was introduced, the only instance I can find of someone noticing this is in a [comment on an unrelated issue](https://github.com/shikijs/shiki/issues/589#issuecomment-2053942608)—however this issue emerges if you wish to display line numbers within a code block using a [CSS counter](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_counter_styles/Using_CSS_counters), which my website does.

Website screenshot, showing an extra blank row after the code block text

### Additional context

This fix only trims the very last newline, as that's the problem one that is received by Shiki from Rehype. I'm assuming that any extra newlines within the code block are the user's own fault (intentional or not), so they probably shouldn't be messed with.


packages/rehype/src/core.ts
```diff
@@ -107,7 +107,10 @@ function rehypeShikiFromHighlighter(
       if (fallbackLanguage && !langs.includes(lang))
         lang = fallbackLanguage
 
-      const code = toString(head)
+      // Replace the single trailing newline that rehype/remark gives us for
+      // some reason. Only gets rid of the very last one, as any extra newlines
+      // are due to the user having blank lines in the code block.
```
antfu, line 112: I'd like to know the "some reason" first before proceeding with this change

Ovyerus: Looks like it gets explicitly added by `mdast-util-to-hast` which is used in `remark-rehype`. It seems to have been like this since the very beginning, so no clue about the reasoning.

https://github.com/syntax-tree/mdast-util-to-hast/blob/f511a93817b131fb73419bf7d24d73a5b8b0f0c2/lib/handlers/code.js#L22

PR #738, opened 2024-08-11
fix: add .nvmrc file

lilpolymath:
### Description

I forked the repo to work on a transformer and I tried to run `pnpm i` with node version 20 and it didn't work. However, node version 18 worked. bv This update locks the repo to a version that currently works with `pnpm`. 

Error attached below.
Terminal Screenshot of installing with node version 20

### Linked Issues
Doesn't apply.

### Additional context

Doesn't apply.


antfu:
I don't mind adding nvmrc - but I don't think your problem is related to your node version (maybe your config) - I am on Node 20 without problem

lilpolymath:
@antfu Alright, thanks. I will double check my installation then.

antfu:
Closing for now as I don't yet want to limit contributors's Node version, as the code is supposed to be runtime agnostic unless proved wrong

PR #739, opened 2024-08-12
feat(core): lighter token explanation with `scopeName`

### LINKED ISSUE #736: Option to exclude themeMatches from explanations

I want an option to exclude `themeMatches` from token explanations, but still include `scopeName`, so that plugins that need scope information can work without the slowdown of including all the themeMatches too

### Suggested solution

In @shiki/core codeToTokensBase could check an additional option to add explanations but skip `themeMatches`

### Alternative

_No response_

### Additional context

I wrote a [shiki transformer for VSCode-style bracket colorization](https://github.com/MichaelMakesGames/shiki-colorized-brackets). It uses the textmate scopes for avoid matching some potential brackets (eg `>` when used as "less than") and to detect embedded languages (eg `ts` in a `svelte` file). It does not use the `themeMatches` on the explanations. I did some benchmarking and noticed that shiki is ~4.7x slower when `includeExplanation` is true, but when I manually edit the code to skip adding `themeMatches` it is only ~2x slower.

### Validations

### Contributes

- [X] If this feature request is accepted, I am willing to submit a PR to fix this issue

antfu:
fix #736

### DIFF
--- a/packages/core/src/code-to-tokens-base.ts
+++ b/packages/core/src/code-to-tokens-base.ts
@@ -125,28 +125,6 @@ function _tokenizeWithTheme(
   let actual: ThemedToken[] = []
   const final: ThemedToken[][] = []
 
-  const themeSettingsSelectors: ThemeSettingsSelectors[] = []
-  if (options.includeExplanation) {
-    for (const setting of theme.settings) {
-      let selectors: string[]
-      switch (typeof setting.scope) {
-        case 'string':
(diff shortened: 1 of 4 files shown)


packages/core/src/code-to-tokens-base.ts
MichaelMakesGames, line 1: The `if` starting on line 129 can be changed to
```ts
if (options.includeExplanation && options.includeExplanation !== 'scopeName') {
```
since the `themeSettingsSelectors` do not end up being used in that case

Issue #660, opened 2024-04-21
`@shikijs/rehype`: support for inline code


jeremt:
### Clear and concise description of the problem

There is no support for highlighting inline code (similarly to [rehype-pretty-code](https://github.com/rehype-pretty/rehype-pretty-code)). It would be nice to be able to highlight something like `console.log('coucou')`.

### Suggested solution

Apply highlighting when there is something like \`console.log('coucou'){:js}\`

### Validations

- [X] Follow our [Code of Conduct](https://github.com/eslint-stylistic/eslint-stylistic/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://eslint.style/contribute/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Contributes

- [ ] If this feature request is accepted, I am willing to submit a PR to fix this issue

antfu:
It would be a good feature to have - however I don't know what syntax we should follow as there is not standard right now. I see people doing

- ``` `code`{lang=js} ```
- ``` `code`{js} ```
- ``` `code`{.js} ```
- ``` `code{:js}` ```

souporserious:
In my own parser, I parse a prefix that works similar to normal triple slash code blocks:
`` `js const a = 'a'` ``

And similarly these can be tagged with a filename:
`` `filename.js const a = 'a'` ``

The presence of the space is used to determine if it needs highlighting, so this wouldn't be highlighted for example:
`` `js` ``

The idea was to bring familiarity with how it's already done in code blocks. However, this does unfortunately look weird in raw markdown since it's embedded in the code snippet.

fuma-nama:
I think \`code{:js}\` will be better since people migrating from `rehype-pretty-code` will need it and it shouldn't break existing transformers (theoretically). This doesn't follow the standard markdown format and I'm not sure if any change is needed to support this feature.

jcayzac:
I've used ``[`some code` lang]`` a lot in another (closed source) project, FWIW.

jcayzac:
Also, just noting that `` `code`{js}`` would be ambiguous in MDX.

fuma-nama:
Working on the PR, it's working very great. Should it be enabled by default?

antfu:
https://github.com/shikijs/shiki/pull/751 get shipped. Free feel to bring more options if you needed as PRs. Closing this issue now to clean up our tracker. Thanks

ryoppippi:
[...]

PR #751, opened 2024-08-20
feat(rehype): support inline codes

fuma-nama:
### Description

Support to highlight inline codes with `code{:language}` syntax.

It isn't enabled by default to avoid problems when people upgrading from older versions of Shiki.
The output is basically same as normal code blocks but with `pre` replaced by a `span` element.

```html
<span class="shiki ...">
<code>...</code>
</span>
```

### Linked Issues

#660 

### Additional context

Should we give additional information to transformers, or allow to pass different options for inline code highlighting?


jcayzac:
Using the proposed syntax `code{:language}` makes the plugin incompatible with MDX.

fuma-nama:
image

Hmm it should work on both markdown, MDX, and probably the markdown syntax of Vitepress too.
```md
`console.log("Hello World"){:js}`
```

packages/rehype/src/core.ts
```diff
@@ -45,15 +52,22 @@ export interface RehypeShikiExtraOptions {
...
-    tree: Root
+    tree: Root,
   ) => Record<string, any> | undefined | null
 
+  /**
+   * Highlight inline code blocks
+   *
+   * @default false
+   */
+  inline?: boolean
```
antfu, line 63:
```suggestion
  inline?: false | 'curly-colon'
```

Something like this to indicate the style? In case we might have multiple ones in the future as https://github.com/shikijs/shiki/issues/660#issuecomment-2069110522

fuma-nama: Brilliant idea, I've updated the docs too

jcayzac:
Is this just for `lang`? How do you pass the rest of the `meta` string? e.g. `js {1,3-4}` when using [`transformerMetaHighlight`](https://shiki.style/packages/transformers#transformermetahighlight)?

fuma-nama:
It doesn't work with transformers, I would just suggest to use code blocks if you want to use them.

jcayzac:
> It doesn't work with transformers, I would just suggest to use code blocks if you want to use them.

Why not making it work?

Using code blocks for inline code isn't really an option.

fuma-nama:
It should be discussed in a separate issue with enough details on how transformers can be re-implemented for inline codes, and it makes no sense to use transformers like line highlight, focus on just an inline code

antfu (approved):

jcayzac:
> it makes no sense to use transformers like line highlight, focus on an inline code.

I use `transformerMetaWordHighlight` in inline code (I pass the `meta` programamtically, using a component, not through rehype).

I don't understand the point about avoiding changes on transformer API being better. That's my point exactly. There's nothing special about inline code.

fuma-nama:
Because this PR is about the Rehype integration of Shiki, you can contribute by opening a PR.

Issue #619, opened 2024-03-09
Change `@shikijs/transformers` range by `+ 1`


richardgill:
### Validations

- [X] Follow our [Code of Conduct](https://github.com/shikijs/shiki/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://github.com/shikijs/shiki/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

### Describe the bug

```
// [!code highlight:2]
console.log('highlighted')
console.log('not highlighted')
```

Ranges on transformations include the current line, to me this was very uninuitive. 

Instead it could be:

```
// [!code highlight:2]
console.log('highlighted')
console.log('highlighted')
```

Currently: `// [!code highlight]` is effectively `// [!code highlight:1]` 
Becomes: `// [!code highlight:0]` 

I guess this is a breaking change, and it is slightly arbitary - so no worries if you disagree :) 

### Reproduction

Included in comment

### Contributes

- [X] I am willing to submit a PR to fix this issue
- [X] I am willing to submit a PR with failing tests

fuma-nama:
Noticed this problem too, it's because the notation highlighter was initially designed as:

```ts
console.log('highlighted') // [!code highlight:2]
console.log('not highlighted')
```

See https://shiki.style/packages/transformers#transformernotationhighlight

So when the comment was added to an empty line, it still start counting from the current line.
The exact problem is that we need to consider the case of empty lines, we still need the approval from maintainers since it should be a breaking change (impacts normal usages).

PR #755, opened 2024-08-22
fix(transformers): Ignore empty lines on notation transformer range

fuma-nama:
### Description

Start counting from next line of the comment if the current line is empty.

For example:

```ts
// [!code highlighted:2]
highlighted
highlighted
```

```ts
highlighted // [!code highlighted:2]
highlighted
```

### Linked Issues

#619 

### Additional context

I made some changes to the design of `createCommentNotationTransformer` so it matches the comment token without a given regex, I've documented it in case for other contributors who doesn't understand it well.

note: forgot to reset local repo after the last PR was merged, apologies for the messy commit history.


fuma-nama:
Also I realized this is a good opportunity to define the standard behaviors of notation transformers.

Like:

**Normal**
```ts
console.log("hello") // [!code xxx] original comment
```

**Multiple transformers**
```ts
console.log("hello") // [!code highlight], [!code world:hello] original comment
```

**impact next line if the current line is empty (breaking?)**
```ts
// [!code highlight], [!code world:hello]
console.log("hello") // original comment
```

**keep empty line (?)**
```ts
// [!code highlight] [!code keep]
console.log("hello") // [!code highlight]
```

it requires some extra work but won't cause a break change except the current proposal of "Ignoring empty lines"

which can fix #589 and other issues related to it.

packages/transformers/src/utils.ts
```diff
@@ -1,12 +1,23 @@
...
+ * @param onMatch function to be called when found a comment in code, remove the comment node if returns `true`.
+ * @param removeEmptyLines remove empty lines below if matched
+ */
 export function createCommentNotationTransformer(
   name: string,
-  regex: RegExp,
   onMatch: (
     this: ShikiTransformerContext,
-    match: string[],
+    commentText: string,
```
antfu, line 20: This API change is a breaking change (this function is a public API), can we wrap another API instead?

fuma-nama: Oops didn't notice that, maybe we can leave a legacy version of the function and add an "experimental" prefix to the new one.

fuma-nama:
@antfu I'm fixing some edge cases and I found one of the test cases:

```js
function hello(indentSize, type) {
  if (indentSize === 4 && type !== 'tab') {
    	console.log('Each next indentation will increase on 4 spaces'); // [!code error] // [!code focus]
  }
}
```

gives the mismatch:

image

I thought it was the problem of my changes, but it seems like the original snapshot was wrong.
The original snapshot didn't apply the error transformer, and my fix accidentally fixed it.

packages/transformers/src/utils.ts
```diff
@@ -1,58 +1,67 @@
 import type { Element, Text } from 'hast'
 import type { ShikiTransformer, ShikiTransformerContext } from 'shiki'
 
-export function createCommentNotationTransformer(
+/**
+ * Regex that matches code comments
+ */
+const regex = /((?:\/\/|\/\*|<!--|[#"']|--|%%|;{1,2}|%{1,2})\s*)(\S.*?)(-->|\*\/|$)/
```
antfu, line 7: I am not sure if we should hard-code the regex, this makes the code less flexiable

packages/transformers/src/utils.ts
```diff
@@ -1,58 +1,67 @@
...
+        const text = last.children[0]
+        if (text.type !== 'text')
+          return
+
+        let deleteComment = false
+
+        const isEmptyLine = line.children.length === 1
+        text.value = text.value.replace(regex, (_, prefix, text, end) => {
+          // no other tokens except the comment
+          const replaced = onMatch.call(this, text, line, last, lines,
```
antfu, line 47: If the second arg `text` is coming from the match, why do we need to change it from `match` to `text`? Doesn't it make the utils less flexiable?

Instead of making a new function as breaking changes, couldn't we just allow `onMatch` to return a string instead of boolean?

fuma-nama: the new `match` here refers to the entire comment (`// [!code] highlight`), but we also want to remove the comment if the comment content becomes empty, e.g. `// `.

so we pass the `text` instead (`[!code] highlight`), the transformer replaces comment content, which allows us to improve the flexibility further, and remove duplicated regex for matching code comments.

As you see, the `createCommentNotationTransformer` now has the algorithm for matching comments. We can improve it to support other comment syntaxes.

For the problem about breaking change, I think we could also create a wrapper over the original function, would like to hear your opinions about this.

packages/transformers/src/utils.ts
fuma-nama: Hmm do you mean to allow passing the entire regex, including the part for matching comments? (like the previous code)

I assume `createCommentNotationTransformer` is for matching comment notations, it sounds reasonable to me to add the regex for matching comment inside the utility. It allows a better optimization, because we know comments are supposed to be the last comment, and we can adjust it depending on the language.

fuma-nama:
I'm doing some experiments on notation transformers, and I found it's possible to redesign the current algorithm of notation transformers to support JSX comments e.g. #770, and optimize performance by reading only the last few tokens of each line.
Parsed comments can be cached per codeblock so chaining multiple notation transformers can lose less performance.

I putted [these changes here](https://github.com/fuma-nama/shiki-transformers), but this introduces breaking change as some backward incompatible changes are made, including *"Ignore empty lines on notation transformer range"* (the aim of this PR).

antfu:
Feel free to always create the PR (if you already work on it), so we can discuss about the implementations.

Sorry I was taking a long day offs, need to catch up and review a lot things

fuma-nama:
Thanks, I'll open a better PR later with more details and explanations. Appreciate your hard work

PR #762, opened 2024-08-30
feat: use `regex` to enhance js engine support

antfu:
https://github.com/antfu/oniguruma-to-js/issues/1


slevithan:
Continuing the discussion from https://github.com/antfu/oniguruma-to-js/issues/1 , here are some basic `regex` plugins that add support for more `oniguruma-to-js` features. I'm curious what level of support for existing TextMate grammars this would give:

```js
import { regex } from 'regex'
import { Context, replaceUnescaped } from 'regex-utilities'

const TABLE_POSIX = {
  alnum: '0-9A-Za-z',
  alpha: 'A-Za-z',
  ascii: '\\0-\\x7F',
  blank: ' \\t',
  cntrl: '\\0-\\x1F\\x7F',
  digit: '0-9',
  graph: '!-~',
  lower: 'a-z',
  print: ' -~',
  punct: '!-/:-@[-`{-~',
  space: ' \\t\\r\\n\\v\\f',
  upper: 'A-Z',
  xdigit: '0-9A-Fa-f',
  word: '\\w',
}

function extensions(str) {
  return str.replace(/(?<slash>\\\\)|\\(?<escapedLiteral>['"`#])|\[(?<neg1>\^?)\[:(?<neg2>\^?)(?<posix>[a-z]+):\]\]/g, (...args) => {
    let groups = args[args.length - 1]
    let {slash, escapedLiteral, neg1, neg2, posix} = groups;
    if (slash) {
      return args[0]
    }
    if (escapedLiteral) {
      return escapedLiteral
    }
    if (posix) {
      let negated = neg1 !== neg2
      let resolved = TABLE_POSIX[posix]
      if (!resolved) {
        throw new Error(`Unknown POSIX class "${posix}"`)
      }
      return `[${negated ? '^' : ''}${resolved}]`
    }
  })
}

function nonCharClassExtensions(str) {
  return replaceUnescaped(str, String.raw`\\(?<escape>[hH])`, ({groups: {escape}}) => {
    return `[${escape === 'H' ? '^' : ''}${TABLE_POSIX.xdigit}]`
  }, Context.DEFAULT)
}

function charClassExtensions(str) {
  return replaceUnescaped(str, String.raw`\\(?<escape>[hH])`, ({groups: {escape}}) => {
    if (escape === 'H') {
      throw new Error('TODO: Add support for \\H in character classes')
    }
    return `${TABLE_POSIX.xdigit}`
  }, Context.CHAR_CLASS)
}

regex({
  flags: 'dg',
  subclass: true,
  plugins: [extensions, nonCharClassExtensions, charClassExtensions],
  unicodeSetsPlugin: null,
  disable: {
    x: true,
    n: true,
    v: true,
  },
})({ raw: [p] })
```

Note that this fixes a couple errors from `oniguruma-to-js`. Namely, it fixes the list of whitespace characters matched by `[[:space:]]`, and it correctly does not support `[:posix:]` syntax outside of character classes (maybe I'm reading the code incorrectly).

antfu:
Thank you! I updated the code based on your suggestions, but unfortunately it seems to still have a lot errors. Meanwhile I am a little bit worried about the complexity here - I am still up to give `regex` a try if it managed to cover more cases that `oniguruma-to-js` couldn't do. But for now I want to move things faster, so I will get back to this a bit later. Thanks

slevithan:
> Thank you! I updated the code based on your suggestions, but unfortunately it seems to still have a lot errors.

I've just edited the code in my last comment to add support for `\A`, `\z`, and escaped literal `\@`. This resolves some but not all of the errors. It will still throw for things like `\G` and `(?i:...)`.

> Meanwhile I am a little bit worried about the complexity here

I've added a new feature in regex-utilities 2.3.0 and used it in the updated example code to significantly simplify. Hopefully this is now easier to follow and makes it clearer how to work with `regex` plugins.

> I am still up to give `regex` a try if it managed to cover more cases that `oniguruma-to-js` couldn't do. But for now I want to move things faster, so I will get back to this a bit later. Thanks

For sure, this makes a lot of sense while you're in the rapid development stage. Even beyond that, though, I think the goals of these two projects might be just far enough apart that you will continue to prefer not to develop on top of `regex`. `oniguruma-to-js` is about doing whatever it takes (and tolerating incomplete solutions) to make the majority of or all TextMate grammars work, whereas `regex` is about modernizing JS regexes with just the most valuable features for readability/safety/power and providing complete/robust solutions for the features it offers. This puts them a bit at odds, even given its plugin system.

So no worries at all if you want to close this. It was interesting/valuable to discuss this with you and see how `regex` might be used in libraries like this that are about cross-flavor compatibility. It's possible that, in the future, `regex` will get more features and more powerful plugins that would make it worthwhile to reexamine.

package.json
```diff
@@ -52,6 +52,8 @@
     "picocolors": "catalog:",
     "pnpm": "catalog:",
     "prettier": "catalog:",
+    "regex": "^4.2.1",
+    "regex-utilities": "^2.3.0",
```
slevithan, line 56: No longer need `regex-utilities` (unless you want to add plugins to `regex`'s `rewrite` function).

scripts/report-engine-js-compat.ts
```diff
@@ -6,9 +6,38 @@ import fs from 'node:fs/promises'
...
+      disable: {
+        n: true,
+        v: true,
+        x: true,
+      },
+    })
+
+    try {
+      return onigurumaToRegexp(rewritten.expression, {
+        flags: rewritten.flags,
```
slevithan, line 31: I'm guessing that lots of the new errors in the `engine-js-compat.md` results are the result of adding the `u` flag here. What happens if you just use `'dgm'` here instead of `rewritten.flags`?

Not using `u` would be a shame because JS needs flag `u` to support syntax like `\u{61}` and `\p{Letter}` (both of which are supported by Oniguruma) and do code point based matching. But maybe such syntax isn't used much in TextMate grammars. (Using `u` while avoiding the errors it produces is certainly possible, but would require doing a few adjustments to the pattern's syntax beforehand.)

package.json
```diff
@@ -52,6 +52,8 @@
     "picocolors": "catalog:",
     "pnpm": "catalog:",
     "prettier": "catalog:",
+    "regex": "^4.2.1",
```
slevithan, line 55: Bump to v4.3.0 (due to my later suggestion to drop flag `u`).

packages/core/package.json
```diff
@@ -72,6 +72,7 @@
   "devDependencies": {
     "hast-util-to-html": "catalog:",
     "oniguruma-to-js": "catalog:",
+    "regex": "^4.2.1",
```
slevithan, line 75: Ditto

docs/references/engine-js-compat.md
```diff
@@ -1,250 +1,250 @@
...
-|                 |                            Count |
-| :-------------- | -------------------------------: |
-| Total Languages |                              213 |
-| Fully Supported | [37](#fully-supported-languages) |
-| Mismatched      |       [2](#mismatched-languages) |
-| Unsupported     |    [174](#unsupported-languages) |
+|                 |                             Count |
+| :-------------- | --------------------------------: |
+| Total Languages |                               213 |
+| Fully Supported | [132](#fully-supported-languages) |
```
antfu, line 14: This gets a lot better, thanks for the updates!

slevithan: Cool. I might not be able to get to it for a little while due to travel, but I'll come back to this and try to identify and fix cases of grammars that used to be okay but aren't anymore. I already notice an issue with `regex`'s support for multiple possessive quantifiers in the same pattern that should be easy to fix.

slevithan:
I've published a fix for use of multiple possessive quantifiers in `regex` 4.3.1. This should allow more grammars to pass.

However, I'm having trouble generating Shiki's `scripts/report-engine-js-compat.json` so I can test this myself.

When I run `pnpm report-engine-js`, I get a bunch of "Sample not found" logs. When I run `npx jiti scripts/report-engine-js-compat.ts` as instructed in `shiki/bench/engines.bench.ts`, I get the following error:

```sh
$ npx jiti scripts/report-engine-js-compat.ts
C:\Users\steve\code\shiki\node_modules\.pnpm\jiti@1.21.6\node_modules\jiti\lib\index.js:2
  throw err; /* ↓ Check stack trace ↓ */
  ^

Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined in C:\Users\steve\code\shiki\packages\core\node_modules\@shikijs\vscode-textmate\package.json
    at exportsNotFound (node:internal/modules/esm/resolve:296:10)
    at packageExportsResolve (node:internal/modules/esm/resolve:586:13)
    at resolveExports (node:internal/modules/cjs/loader:639:36)
    at Module._findPath (node:internal/modules/cjs/loader:747:31)
    at Module._resolveFilename (node:internal/modules/cjs/loader:1234:27)
    at Function.resolve (node:internal/modules/helpers:145:19)
    at _resolve (C:\Users\steve\code\shiki\node_modules\.pnpm\jiti@1.21.6\node_modules\jiti\dist\jiti.js:1:241814)
    at jiti (C:\Users\steve\code\shiki\node_modules\.pnpm\jiti@1.21.6\node_modules\jiti\dist\jiti.js:1:244531)
    at C:\Users\steve\code\shiki\packages\core\dist\index.mjs:2:23
    at evalModule (C:\Users\steve\code\shiki\node_modules\.pnpm\jiti@1.21.6\node_modules\jiti\dist\jiti.js:1:247313) {
  code: 'ERR_PACKAGE_PATH_NOT_EXPORTED'
}

Node.js v22.8.0
```

slevithan:
Okay, I've run the `report-engine-js` script locally, and here's what I've found.

- There were two possessive quantifier related bugs in `regex` v4.3.0 that were being triggered, that are now fixed in v4.3.2. This significantly raises the number of grammars passing.
- I'm not sure what the purpose of the following code is in the latest version of this PR:

```js
    pattern = pattern
      .replace(/\\￿/g, '\\G')
      .replace(/\(\{\)/g, '(\\{)')
```

Removing this increases the number of grammars passing, whether using `regex` or not. Without it, using `regex` gives the following results:

```
| Fully Supported | [172](#fully-supported-languages) |
| Mismatched | [24](#mismatched-languages) |
| Unsupported | [17](#unsupported-languages) |
```

This is still a bit lower than when not using `regex`, but I think at this point it's worth adding anyway, because the cases where it is failing are cases where `oniguruma-to-js` is getting things wrong even though it passes these tests!

The culprits are the following errors that `regex` is throwing:

- "Subroutine `\g<name>` followed a recursive reference"
- "Invalid named capture referenced by subroutine"
  - For example, given `\g<1>` or `\g<-1>`.

These are subroutines that Oniguruma supports but `regex` doesn't, as explained in https://github.com/antfu/oniguruma-to-js/issues/1#issuecomment-2329888133:

> - `regex` doesn't support the following subroutine features (they will throw if encountered):
>   - Subroutines that reference groups by absolute or relative number via `\g<N>`.
>   - Recursive subroutines. These can't be emulated using JS regexes.

Even though `oniguruma-to-js` doesn't throw on these patterns and the tests are not catching the resulting problems (given the specific grammar samples being tested), all of these patterns are being interpreted incorrectly. So IMO it is better that `regex` is throwing for them. What is happening currently (without `regex`) is that, since not using flag `u` or `v`, the `\g` is being interpreted as a literal 'g' (since `\g` is not a known escape), and therefore e.g. `\g<1>` is being treated as equivalent to `g<1>`, which matches only the four literal characters `g`, `<`, `1`, `>`, in sequence.

Looking at the remaining failing cases beyond the issues described above, the main culprits seem to be:

- Nested character classes `[...[...]]` and character class intersection `[...&&...]`.
  - These would be supported if using JS's flag `v`, but then you'd have to deal with flag `v`'s strict errors and different escaping rules, which would cause a LOT more errors unless you first transpiled many more things to match flag `v`'s stricter rules.
  - Technically these features don't cause errors when not using flag `u`/`v`, but sometimes you get other errors as a side effect like "Range out of order in character class" if unescaped `-` is the first char in the nested class. And in cases where it doesn't throw, it is currently always making the pattern non-equivalent (so they match the wrong things) even though this might not be caught by the specific grammar samples being tested.
- Unicode categories. Ex: `\p{L}`, `\P{Mn}`, etc.
  - These would be supported if using JS's flag `u` or `v`, but again you'd then have to deal with their strict errors and different escaping rules. When not using these flags, they don't throw an error, but they are always matching the wrong things (noticed in quite a few regexes).

Here's what I'd personally do:

- Bump to `regex` 4.3.2.
- Remove the `pattern = pattern.replace...` line I mentioned at the top of this comment, unless it is serving some important purpose.
- Land the PR.

Even though the reported number of "supported" grammars would go down slightly (to 172), this should only affect grammars that are already not actually working correctly. And the number of *actually* supported grammars would go up *significantly*, since:

1. It should not be adding errors for any regexes that are *actually* working as intended.
2. New grammars (that use atomic groups) will gain support.
3. Many grammars currently "supported" that are not actually working correctly would gain *correct* support. This will be true for any grammars that use possessive quantifiers (which `oniguruma-to-js` currently neuters) and subroutines referencing named groups (which, since you're not using flag `u` or `v`, don't throw but never match what's intended).

Then, after landing, I'd personally recommend doing additional work in `oniguruma-to-js` to transpile non-strict regex syntax into strict `v` mode syntax, and use the resulting patterns with flag `v`. This work is potentially a bit tricky (and would limit support to Node.js 20 and 2023-era browsers), but it would add native support for nested character classes, set intersection/subtraction, Unicode categories, etc. (all non-trivial to emulate, and emulated versions would not be as fast as the native support). I'd also be happy to help with this work, if desired. And after this work, more grammars would be supported than currently are, since these advanced Oniguruma features would just work and run as native JS regex features (additionally, the options used with `regex`'s `rewrite` function could then be simplified).

antfu:
Thanks a lot for your deep investigation, that looks very promising!

> This is still slightly lower than the number reported when not using regex, but the report is misrepresenting the facts based on the limited grammar samples, and in fact the number of correctly supported grammars is now significantly higher than before.

Yes, I totally agree with that. It's surely better to surface the errors instead of silently eating them.

With the current result, I am more than happy to land it, and then we can iterate as we go. However, as the JS engine is mostly designed to work on browsers, the bundle size is something we need to be concerned about. If `regex` can do the work much better and correctly than `oniguruma-to-js`, I'd be more than happy to deprecate it and use `regex` exclusively.

The current test runs with `regex -> oniguruma-to-js`, where if I remove the `oniguruma-to-js` part, it will throw many more syntax errors. Is it possible we port those fixes to `regex` so `shiki` can solely deps on `regex` then?

> This work is potentially a bit tricky (and would limit support to Node.js 20 and 2023-era browsers

For Shiki, in this particular case, I think it's acceptable as we consider this to be an advanced usage. If it's tricky to provide back-compact, we can just document the minimal runtime requirement for using the feature. 

Thanks so much for your work and help!

antfu:
I'll merge this for now, let's iterate it on main banch.

slevithan:
Amazing to see this merged. 🎉🙌

> as the JS engine is mostly designed to work on browsers, the bundle size is something we need to be concerned about. If `regex` can do the work much better and correctly than `oniguruma-to-js`, I'd be more than happy to deprecate it and use `regex` exclusively.

If I/we create a new library that extends `regex`, I could definitely help improve Oniguruma translation further and with greater accuracy.

It wouldn’t make sense to add extensive Oniguruma translation to the base `regex` lib since `regex`’s syntax/flags/features are not based on Oniguruma. Its much more based on PCRE, and even then it is not trying to support all PCRE features — it's only adding a minimal set of features that are most relevant for elevating native JS regexes to be one of the best and most readable/modern regex flavors in its own right.

**But**, it absolutely makes sense to me to create a new library (tentative name: `textmate-grammar-regex`; or maybe `regex-plugins` which would bundle a bunch of plugins under an "Oniguruma" flavor set) that builds on top of `regex` and can take over responsibility from `oniguruma-to-js` (since you mentioned you're okay with deprecating it), with Shiki as a first-class client library. Then, at the same time, I could offer new exports in the base `regex` lib that make it much more tree shake-able when used in this way, so that the new lib can be as lightweight as possible.

I have a bunch of ideas for this, but I won't be able to start development for it until sometime next week. I’d be happy to add you as a maintainer for this new project (totally up to you!) after the basics are set up.

What do you think? Does this seem like a decent plan?

antfu:
Sounds great, looking forward to it! (and take your time ofc). Thanks!

slevithan:
Cool. I think it will be a fun project so I'm looking forward to it, too.

> > This work is potentially a bit tricky (and would limit support to Node.js 20 and 2023-era browsers
> 
> For Shiki, in this particular case, I think it's acceptable as we consider this to be an advanced usage. If it's tricky to provide back-compact, we can just document the minimal runtime requirement for using the feature.

Okay, in that case I'm going to require native flag `v` support in the new lib. Transpiling the new features from flags `u`/`v` is indeed tricky, but there's a great existing lib that already does this: [regexpu-core](https://github.com/mathiasbynens/regexpu-core) (which is used under the hood by Babel to transpile `u`/`v`). The problem, and the reason I wouldn't recommend using it, is that it would make the resulting package very large (you mentioned that small bundle size is important) because it includes a ton of necessary Unicode character data. It would also make the resulting regexes at least somewhat slower (and much larger) than just relying on the native versions of these features that you get from flag `v`.

slevithan:
Heads up that I've indeed started work a couple days ago on a much more robust and accurate Oniguruma to JS transpiler. It will probably have ~100 big and small fixes/improvements (tons of edge cases to investigate and deal with!) compared to the current lib. But it's a moderately big project as I need to write a full compiler, etc. to do it justice while keeping package size in mind. So it will take some time.

PR #767, opened 2024-09-02
feat(shiki-monaco): add `langMappings` option

nag5000:
### Description

There is currently no way to apply tsx syntax highlighting from shiki without losing the built-in IntelliSense for the typescript language in monaco. There is no `tsx` language in monaco - tsx is covered by `typescript` language. The guide (https://shiki.style/packages/monaco) suggests doing `monaco.languages.register({ id: ... })`, but by doing so for `id: 'tsx'`, we will lack the built-in IntelliSense for typescript.

So I added an option to map between shiki and monaco languages. For tsx it looks like this:

```
 const shikiHighlighter = await createHighlighter({
   themes,
   langs: ['tsx'],
 })

 shikiToMonaco(shikiHighlighter, monaco, {
   langMappings: { tsx: 'typescript' }
 })
```

We can then use the built-in typescript language in monaco, with TSX syntax highlighting from shiki:

```
monaco.editor.createModel(
   ...,
   'typescript',
   monaco.Uri.parse('file:///index.tsx')
 )
```

### DIFF
--- a/packages/monaco/src/index.ts
+++ b/packages/monaco/src/index.ts
@@ -18,6 +18,12 @@ export interface ShikiToMonacoOptions {
    * @default 500
    */
   tokenizeTimeLimit?: number
+  /**
+   * Mappings from Shiki language identifiers to Monaco language identifiers.
+   *
+   * @default {}
+   */
+  langMappings?: Record<string, string>
 }
 
 export function textmateThemeToMonacoTheme(theme: ThemeRegistrationResolved): MonacoTheme {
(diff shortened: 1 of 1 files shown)


antfu:
Would this option on the highlighter work? https://github.com/shikijs/shiki/blob/aec5f19f2a5c1799f411d680bb9bddd36d4dd200/packages/core/src/types/options.ts#L22-L26

nag5000:
Oh, I missed that option. Yeah, it seems like this option would work. Thanks for the help!

PR #787, opened 2024-09-21
feat(rehype): support lazy load languages

### LINKED ISSUE #786: Lazy load languages on Rehype Shiki

I noticed that Rehype Shiki doesn't actually use async on `rehypeShikiFromHighlighter`, hence it cannot load languages on-demand, making it unsuitable for browser usage.

I know it was to support `processSync` usage for unified processor, but mainly two reasons:
1. async is very common nowadays (in both browser/node.js)
2. it's possible without losing sync support

### Suggested solution

Support a `lazy` option to enable loading languages lazily.
When enabled, the plugin returns a promise instead of a sync call, so inside the rehype plugin we can use promise to load languages with `highligher.loadLanguage`.

### Alternative

_No response_

### Additional context

_No response_

### Validations

### Contributes

- [X] If this feature request is accepted, I am willing to submit a PR to fix this issue

fuma-nama:
### Description

Support loading languages on demand with Rehype Shiki `lazy` option.
When disabled (default), the plugin is compatible with sync usage.

With `lazy` enabled, it loads the language when needed using async `highlighter.loadLanguage`

### Linked Issues

fix #786 

### Additional context

I noticed the amount of duplicated code increase as I try to support sync usage, so I made a "handler" layer.

the handler extracts information from HAST node and tells highlighter how to highlight the node, the logic for highlighting the code is kept in the original `core.ts`.


packages/rehype/src/core.ts
```diff
@@ -69,7 +57,7 @@ function rehypeShikiFromHighlighter(
         {
           name: 'rehype-shiki:code-language-class',
           code(node) {
-            this.addClassToHast(node, `${languagePrefix}${lang}`)
+            this.addClassToHast(node, `language-${lang}`)
```
antfu, line 60: Why changing this? The variable was more like a TODO trying to make it configurable

packages/rehype/test/core.test.ts
```diff
@@ -38,6 +38,31 @@ it('run', async () => {
   expect(file.toString()).toMatchFileSnapshot('./fixtures/a.core.out.html')
 })
 
+it('run with lazy', async () => {
+  const highlighter = await createHighlighter({
+    themes: [
+      'vitesse-light',
+    ],
+    langs: [],
```
antfu, line 46: It doesn't seem to work. The queue resolving order seems to be wrong.

fuma-nama: Sorry for mistakes, fixed the bug

PR #798, opened 2024-10-04
feat(vitepress-twoslash): cache type informations to improve performance

nakasyou:
### Description

VitePress Twoslash takes a long time to render large Types.
This PR solves part of this.

This PR can cache type information with the following code:
```ts
export default defineConfig({
  markdown: {
    codeTransformers: [
      transformerTwoslash({
        typesCacheDir: './my-custom-cache-dir'
      })
    ]
  }
})
```
This will cache the type information in the specified directory.
Use a cache such as GitHub Actions with this to speed up your builds

### Linked Issues

Nothing

### Additional context

ElysiaJS appears to have cut back on Twoslash because it took 30 minutes to build the documentation. This PR may resolve it.
https://x.com/saltyAom/status/1842140132897615963


nakasyou:
Hi @antfu, thank you for commenting.

You mean
```ts
const typesCache = createFileSystemTypesCache(options)

transformerTwoslash({
  typesCache
})
```
, right?

antfu:
Yes, something like that

nakasyou:
@antfu could you check this?

kravetsone:
Great this interesting Q

jasonkuhrt:
This is exciting! I made an issue for Graffle to look more into this when I have time.

When it comes to Vitepress, in my mind it feels like this caching stuff should be the default, and presumably almost no one will need to opt out of it, but if they need to then sure it should be possible.

So, in other words I'm hoping that a future vitepress release will just make everything faster without any effort from user needed by default. Curious to see how it will actually land.

Thanks all!

Issue #734, opened 2024-08-06
Make grammar state working with `themes` instead of just `theme`


xsjcTony:
### Clear and concise description of the problem

Currently it's not allowing to use `themes` when passing grammar state into `codeToHtml`

Either way it's not going to work no matter I specify theme in `.getLastGrammarState` or not. Seems it defaults to the first loaded theme.
See the below example.

Also, seems the grammar state is not in the documentation, so I have to follow the guide in the PR and from TS types.

e.g.
```ts
const highlighter = await getSingletonHighlighterCore({
  themes: [
    import('shiki/themes/vitesse-light.mjs'),
    import('shiki/themes/one-dark-pro.mjs'),
  ],
  langs: [
    import('shiki/langs/typescript.mjs'),
  ],
  loadWasm: getWasmInstance,
});

export const stateTSTypeAnnotation = highlighter
  .getLastGrammarState('let a:', { lang: 'ts' });
// or with
  .getLastGrammarState('let a:', { lang: 'ts', theme: 'one-dark-pro' });
```
```ts
function renderType(type: string): string {
  return codeToHtml(
    type,
    {
      lang: 'ts',
      // theme: 'one-dark-pro', // only this is going to work if I specify `theme` in `.getLastGrammarState()` above
      themes: {
        light: 'vitesse-light',
        dark: 'one-dark-pro',
      },
      grammarState: stateTSTypeAnnotation,
      structure: 'inline',
    },
  );
}
```
![image](https://github.com/user-attachments/assets/d2831790-b388-4872-8c3c-0b2ae912a665)
![image](https://github.com/user-attachments/assets/58fde7e4-fc1f-41e5-bdca-c5e692fc17c5)

### Suggested solution

allow `themes` in `.getLastGrammarState()`, or simply remove `theme` from the function.
I've no idea how it's working under the hood, but it's a bit weird to me since we are passing the grammar state to the render function, where `theme` or `themes` has to be provided, then why do we still need to pass it to grammar state?

### Validations

- [X] Follow our [Code of Conduct](https://github.com/eslint-stylistic/eslint-stylistic/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://eslint.style/contribute/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Contributes

- [ ] If this feature request is accepted, I am willing to submit a PR to fix this issue

uxiew:
[...]

PR #804, opened 2024-10-07
feat: store grammar state in weakmap

antfu:
In https://github.com/shikijs/shiki/pull/712, we introduced GrammarState to resume the highlighting from the middle way. It solves the problem of customizing grammar context, but wasn't very useful for passable highlighting (for example, in a text editor, instead of highlighting the full context again over again, we can reuse the result from the previously highlighted lines) - because users need to run the parsing twice to get the `GrammarState` object. The existing API wasn't designed to be able to return extra property, making it hard to return the `GrammarState` in one-go, without breaking changes.

This PR overcharges the `shiki.getLastGrammarState()` function to be able to accept the return Hast `Root` or `ThemedToken[][]` to retrieve the `GrammarState` stored in a internal WeakMap. Make the usage more performent.

Along the way, we also made `GrammarState` work with multiple themes by storing multiple internal state for each theme.

The usage would be like:

```ts
const shiki = await getHighlighter(/*...*/)

const hast = shiki.codeToHast(/* ... */)

const grammarState = shiki.getLastGrammarState(hash) // pass the hast instead of code
```

close #803

### DIFF
--- a/packages/core/src/constructors/bundle-factory.ts
+++ b/packages/core/src/constructors/bundle-factory.ts
@@ -190,7 +190,9 @@ export interface ShorthandsBundle<L extends string, T extends string> {
    * Shorthand for `getLastGrammarState` with auto-loaded theme and language.
    * A singleton highlighter it maintained internally.
    */
-  getLastGrammarState: (code: string, options: CodeToTokensBaseOptions<L, T>) => Promise<GrammarState>
+  getLastGrammarState:
+    | ((element: ThemedToken[][] | Root) => GrammarState)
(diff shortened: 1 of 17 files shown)


xsjcTony:
@antfu This seems to close https://github.com/shikijs/shiki/issues/734 as well.
But instead of using `highlighter.getLastGrammarState`, I'll need to use `highlighter.codeToTokens()` and `tokens.grammarState` instead.
Not sure if this is intended or did we just miss to change `getLastGrammarState()`?
Anyway at least there's a workaround now for #734, so I guess we can close that one as well?

antfu:
Oh yeah, great to hear!

PR #807, opened 2024-10-11
refactor(cli): code-to-ansi

Simon-He95:

### DIFF
--- a/packages/cli/src/code-to-ansi.ts
+++ b/packages/cli/src/code-to-ansi.ts
@@ -23,7 +23,7 @@ export async function codeToANSI(code: string, lang: BundledLanguage, theme: Bun
         text = c.hex(hexApplyAlpha(color, themeReg.type))(text)
       if (token.fontStyle === FontStyle.Bold)
         text = c.bold(text)
-      if (token.fontStyle === FontStyle.Italic)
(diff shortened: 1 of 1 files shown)


antfu:
Do you have an issue regarding this? I guess that bold and italic can exist at the same time

Simon-He95:
> Do you have an issue regarding this? I guess that bold and italic can exist at the same time您对此有什么问题吗？我想粗体和斜体可以同时存在

`token.fontStyle` is impossible to satisfy both `FontStyle.Bold` and `FontStyle.Italic` at the same time

antfu:
Actually it's a bug in another way: we need to use AND to do bitwis comparison:

https://github.com/shikijs/shiki/blob/2d58f7ff7bd642682e93b577d9b68bede8548ef8/packages/core/src/utils.ts#L202-L209

Simon-He95:
> Actually it's a bug in another way: we need to use AND to do bitwis comparison:实际上，从另一个方面来说，这是一个错误：我们需要使用 AND 来做 bitwis 比较：
> 
> https://github.com/shikijs/shiki/blob/2d58f7ff7bd642682e93b577d9b68bede8548ef8/packages/core/src/utils.ts#L202-L209

My mistake. I ignored text as the next parameter.

Issue #629, opened 2024-03-18
`@shikijs/rehype`: keep hast element `data` and `properties`


remcohaszing:
### Clear and concise description of the problem

With `remark-rehype` / MDX, the following markdown:

````md
```js fileName="hello.js"
console.log('hello')
```
````

generates the following hast:

```js
{
  type: 'element',
  tagName: 'pre',
  properties: {},
  children: [
    {
      type: 'element',
      tagName: 'code',
      properties: {
        className: 'language-js'
      },
      data: {
        meta: 'fileName="hello.js'
      },
      children: [
        {
          type: 'text',
          value: "console.log('hello')",
        }
      ]
    }
  ]
}
```

Given this has, Shiki replaces the `pre` element with a newly generated `pre` element. In this process the node’s `data` and `properties` are lost. It would be nice if Shiki keeps this information, so that other rehype plugins, such as [`rehype-mdx-code-props`](https://github.com/remcohaszing/rehype-mdx-code-props) can use it. Likewise it could even be useful to preserve positional information, which is used for example by React devtools.

### Suggested solution

It looks like some logic is needed in https://github.com/shikijs/shiki/blob/main/packages/core/src/code-to-hast.ts to be able to handle an existing `pre` and `code` element.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/eslint-stylistic/eslint-stylistic/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://eslint.style/contribute/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Contributes

- [X] If this feature request is accepted, I am willing to submit a PR to fix this issue

antfu:
Sure, PR welcome!

FlatMapIO:
It seems that the copy button cannot be implemented without the data 😇

karlhorky:
I was also looking for the meta information from the code fence (the `fileName="hello.js"` part), and I found that there is code that already handles the `node.data.meta` to some degree:

https://github.com/shikijs/shiki/blob/4c1de346d6e373cef3599474e2d8d35d2d98727e/packages/rehype/src/core.ts#L117-L128

But it only parses the meta string and passes the props if a custom `parseMetaString` is specified, as shown in the tests:

https://github.com/shikijs/shiki/blob/4c1de346d6e373cef3599474e2d8d35d2d98727e/packages/rehype/test/index.test.ts#L21-L28

[...]

PR #816, opened 2024-10-24
feat(core): codeToHast supports generating lang-xx on code element

zhangmo8:
### Description

`codeToHast` supports generating language class on `code` element

### Linked Issues

#629 

### Additional context

In this #629 , he also hopes to support `meta data`. I found that there are already relevant data in `codeToHast`, but I observed that the relevant data has been filtered out, so I don't know how to deal with it appropriately. Therefore, there is no relevant logic for now in this PR.

image

https://github.com/shikijs/shiki/blob/f76a371dbc2752cba341023df00ebfe9b66cb3f6/packages/core/src/highlight/code-to-hast.ts#L113-L116

### DIFF
--- a/packages/core/src/highlight/code-to-hast.ts
+++ b/packages/core/src/highlight/code-to-hast.ts
@@ -34,7 +34,9 @@ export function codeToHast(
 ): Root {
   let input = code
 
-  for (const transformer of getTransformers(options))
+  const transformers = getTransformers(options)
+
+  for (const transformer of transformers)
     input = transformer.preprocess?.call(transformerContext, input, options) || input
 
   let {
@@ -62,7 +64,7 @@ export function codeToHast(
     },
   }
 
-  for (const transformer of getTransformers(options))
+  for (const transformer of transformers)
     tokens = transformer.tokens?.call(contextSource, tokens) || tokens
 
   return tokensToHast(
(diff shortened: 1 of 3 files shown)


packages/core/src/highlight/code-to-hast.ts
```diff
@@ -122,7 +133,11 @@ export function tokensToHast(
   let codeNode: Element = {
     type: 'element',
     tagName: 'code',
-    properties: {},
+    properties: addLanguageClass
+      ? {
+          class: `${options.lang ? `language-${lang}` : ''}`,
```
antfu, line 138: I am not sure if this is worth it to add to the core, also feels a bit too specific. It feels like it can be done with a transformer isn't it?

zhangmo8: You mean like this? https://github.com/shikijs/shiki/blob/main/packages/transformers/src/transformers/transformer-meta-highlight.ts

zhangmo8: I think this method is more suitable than transformer. Although the class rendered to dom can be unified in the end, the information of lang will be lost in ast.

antfu: What kind of information do you want? You can get the language via `this.options.lang`.

My concern is that `language-*` is just a common practice, not a standard - what if some users want `lang-*` or `shiki-lang-*` to anything? We can't have an option for every single need - that why transformer exists

zhangmo8: Okay, maybe using transformer is a better choice.

PR #818, opened 2024-10-25
feat(transformers): language class transformer

zhangmo8:
### Description

 language class transformer

The outer `<code>` tag is modified: `<code class="prefixClass-lang">`

### Additional context

Link #816


antfu:
Thanks! But I meant to have it on the userland more, as this is not complicated to implement (3 lines of code). On the other hand, to have it as a built-in transformer would increase our maintenance scope that doesn't hide much complexity from the users. I am not very sure about that.

PR #821, opened 2024-10-29
feat(rehype): Add fallback language support for lazy loading in rehype-shiki

rocklau:
### Description
```ts
 /**
   * The fallback language to use when specified language is not loaded
   *
   * Ignored if `lazy` is enabled
   */
  fallbackLanguage?: string
```
Previously, unsupported languages were ignored if `lazy` was enabled. 

This update assigns a fallback language when the requested language isn’t found, 
ensuring the fallback language is enabled.
 

### Linked Issues
N/A

### DIFF
--- a/packages/rehype/src/core.ts
+++ b/packages/rehype/src/core.ts
@@ -6,6 +6,7 @@ import type { Root } from 'hast'
 import type { Transformer } from 'unified'
 import type { RehypeShikiHandler } from './handlers'
 import type { RehypeShikiCoreOptions } from './types'
+import { bundledLanguages } from 'shiki'
 import { isSpecialLang } from 'shiki/core'
 import { visit } from 'unist-util-visit'
 import { InlineCodeHandlers, PreHandler } from './handlers'
@@ -97,6 +98,9 @@ function rehypeShikiFromHighlighter(
         return lang
 
       if (lazy) {
+        if (fallbackLanguage && !(Object.keys(bundledLanguages) as string[]).includes(lang)) {
+          lang = fallbackLanguage as string
+        }
         languageQueue.push(lang)
         return lang
       }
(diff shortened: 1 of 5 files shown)


packages/rehype/src/core.ts
```diff
@@ -97,6 +98,9 @@ function rehypeShikiFromHighlighter(
         return lang
 
       if (lazy) {
+        if (fallbackLanguage && !(Object.keys(bundledLanguages) as string[]).includes(lang)) {
```
fuma-nama, line 101: how about languages that's not included by default, but specified on user land (e.g. custom lang/grammar)?

packages/rehype/src/types.ts
```diff
@@ -28,7 +28,6 @@ export interface RehypeShikiExtraOptions {
   /**
    * The fallback language to use when specified language is not loaded
```
fuma-nama, line 29: I think a more precise description is "fallback language to use when specified language is not loaded or unavailable". With `lazy` enabled, the language will always be loaded, but it might be unavailable (failed to load). and here's where the `fallbackLanguage` is being used.

packages/rehype/src/core.ts
rocklau: Yeah, For now these situations are not considered. May be check is or not include  LanguageRegistration[]

packages/rehype/src/core.ts
```diff
@@ -6,6 +6,7 @@ import type { Root } from 'hast'
 import type { Transformer } from 'unified'
 import type { RehypeShikiHandler } from './handlers'
 import type { RehypeShikiCoreOptions } from './types'
+import { bundledLanguages } from 'shiki'
```
antfu, line 9: Core module should not include `bundledLanguages`

antfu (changes requested):

fuma-nama:
I personally think it is not very suitable for `lazy` because the functionality of fallback language was to use another language when specified language of codeblock isn't pre-loaded.

For this to work with `lazy`, you have to watch for error in `loadLanguages`, so `fallbackLanguage` will be used as an error fallback. Given that `loadLanguages` throws error for unknown language, it will work similarly.

Issue #303, opened 2022-02-22
Bracket pair colorization


theoludwig:
Hey! :wave:

Thanks for this awesome package, I can finally use VSCode default Dark+ theme syntax highlighting on my blog. :smile:
I'm wondering if [Bracket pair colorization](https://code.visualstudio.com/blogs/2021/09/29/bracket-pair-colorization) could be added to Shiki? It has been added to VSCode natively since v1.60 with `settings.json`: `"editor.bracketPairColorization.enabled": true`.

### With VSCode:

![image](https://user-images.githubusercontent.com/25207499/155231222-6fec990e-8758-49ed-910f-fa426e4fd14b.png)

### With Shiki

![image](https://user-images.githubusercontent.com/25207499/155231320-3cc65428-43d9-4b84-b35c-c53afbf87de6.png)

muenzpraeger:
That should be in [now with 0.12.0](https://github.com/shikijs/shiki/releases/tag/v0.12.0).

Divlo:
> That should be in [now with 0.12.0](https://github.com/shikijs/shiki/releases/tag/v0.12.0).

How do you enable it? It seems like it is not enabled by default? I updated to `shiki@0.12.1`, and there is still no bracket pair colorization. @muenzpraeger

muenzpraeger:
@Divlo VSCode itself has built-in bracket-pair colorization (based on runtime inspection of the code). If that should be replicated in a non-IDE situation like here with Shiki the `tmLangauge` definition as well as the the used theme have to provide that information.

octref:
This requires extracting the VS Code bracket pair colorization logic out and bringing it to shiki. I don't quite have the bandwidth to do it right now.

hunghg255:
@theoludwig I tried to create a transformer, you can refer to my code [transformerBracketPairColor](https://gist.github.com/hunghg255/0fa614ffb8dd0107d39d90a4f131c810)
Hope can solve your problem

<img width="498" alt="Screenshot 2024-02-14 at 11 10 10" src="https://github.com/shikijs/shiki/assets/42096908/db6400f4-4dcc-4025-bcec-c62f3d33b15a">

MichaelMakesGames:
Hey all, I just published a plugin for Shiki that can do this: [GitHub](https://jsr.io/@michael-makes/shiki-colorized-brackets) / [JSR](https://jsr.io/@michael-makes/shiki-colorized-brackets)

It is used as a tranformer, like this:
```ts
[...]

PR #831, opened 2024-11-08
feat: port `@michael-makes/shiki-colorized-brackets` JSR package

MichaelMakesGames:
### Description

There is a long-standing feature request to add VSCode-style bracket pair colorization to Shiki. I created a transformer that accomplishes this, and after discussion in the linked issue, it sounds like there's interest in incorporating it as an official plugin.

### Linked Issues

fix https://github.com/shikijs/shiki/issues/303

### Additional context

This is mostly a straight port of my package, https://github.com/MichaelMakesGames/shiki-colorized-brackets. I've made some efforts to align with the styles and patterns I observed elsewhere in the Shiki repo, but let me know if there are other improvements I can make.

I'd like to call out that the way this splits tokens apart feels a little hacky, mostly just doing some basic text search. I investigated another route that used the token explanations, but that had issues highlighting mismatched brackets. If that can be worked around, that would be a more robust solution.

Question regarding the docs: currently this transformer isn't enabled in the docs. Would we want this enabled there? Perhaps only for the page for this package?

Another note: I listed myself as the author of the package. I noticed all the other packages list @antfu. I'm fine changing that.


antfu:
Awesome! Thanks a lot! Please give me some time to properly review.

> splits tokens apart feels a little hacky

Do you consider using https://shiki.style/guide/decorations? It should handle the complexity of token splitting, where you just need to provide the positions.

> Would we want this enabled there? Perhaps only for the page for this package?

Yeah I would vote to enable it only for the page of that package. You can refer to how we do that with theme or other transformers.

> I listed myself as the author of the package.

Surely! It totally make sense to have your name on that package!

MichaelMakesGames:
> Yeah I would vote to enable it only for the page of that package

This is done in the latest commit. I added an `explicitTrigger` option, similar to twoslash, so the plugin can be enabled only for certain code blocks based on the meta string

antfu (approved):

antfu:
Thanks a lot! Let's have it and see how it works

Issue #828, opened 2024-11-07
Switch JS engine to Oniguruma-To-ES for much improved emulation


slevithan:
I've just released [Oniguruma-To-ES](https://github.com/slevithan/oniguruma-to-es), based on our previous conversation #762. There's room for improvement, but it's already very solid, and it includes detailed documentation (and a [demo page](https://slevithan.github.io/oniguruma-to-es/demo/)). Adopting it would put Shiki on a solid path to being able to use the JS engine by default in the future if you want to, and not have to label it as experimental.

## Compatibility report

The JS regex engine [compatibility report](https://github.com/shikijs/shiki/blob/main/docs/references/engine-js-compat.md) currently says 176 supported, 23 mismatched, and 15 unsupported. Running it with Oniguruma-To-ES v0.1 with "loose" `accuracy` gives: 178 supported, 18 mismatched, and 18 unsupported (but see below for why this should be higher).

```js
return toRegExp(pattern, {
  accuracy: 'loose',
  global: true,
  hasIndices: true,
  tmGrammar: true,
})
```

(Note: ["loose"](https://github.com/slevithan/oniguruma-to-es#accuracy) makes its `\G` handling simply set `sticky` when it doesn't have an existing emulation strategy for the specific way `\G` is used, like oniguruma-to-js does. But even when this is on, Oniguruma-To-ES has a lot more strategies for precise `\G` emulation than the current JS engine.)

However, one problem with these numbers is that Oniguruma-To-ES is throwing for patterns with genuine Oniguruma errors (they include errors by their authors) that Shiki's current JS engine is unable to detect problems for (because they're not errors in JS regexes). In these cases, the current engine is just passing regexes that are invalid in Oniguruma through, whereas appropriate Oniguruma-To-ES errors are preventing this from happening. But vscode-textmate silently fails when the real Oniguruma throws an error! So to account for this difference, let's try the report with `forgiving` on:

- **Current engine**: **178** supported, 36 mismatched, 0 unsupported.
- **Oniguruma-To-ES**: **183** supported, 31 mismatched, 0 unsupported.

**In fact, this is a more accurate representation of results.** Because TextMate grammars fail silently for Oniguruma regexes that error, I think it would be appropriate to change the JS engine to *always* use `forgiving` mode, and maybe rename it as something like `failOnError` with default `false`. The "unsupported" and "mismatched" lists should be *merged*, although it would be good to to keep the `failOnError` option around for development (to help identify potential opportunities for improvement).

[...]

PR #832, opened 2024-11-10
feat: Use Oniguruma-To-ES in the JS engine (#828)

slevithan:
### Description

See #828. This changes the JS engine to use [Oniguruma-To-ES](https://github.com/slevithan/oniguruma-to-es) for much improved emulation (beyond what the numbers say) and sets a new foundation to continue improving grammar support in the JS engine.

### Additional context

I don't really understand what's going on with the `packages/engine-javascript/test` tests (which I didn't modify in this PR except for deleting `utils.test.ts` since it was testing a now-deleted file), so it's possible some stuff in there should also be deleted or modified after this lands. `pnpm test` passes.

The `simulation` option in `packages/engine-javascript/src/index.ts` is probably not appropriately named anymore now that most of its behavior has been removed. I've updated the comment for it but didn't change the name (will leave for @antfu).

### DIFF
--- a/packages/engine-javascript/test/compare.test.ts
+++ b/packages/engine-javascript/test/compare.test.ts
@@ -161,7 +161,10 @@ describe.skipIf(
     const run = c.c.skip ? it.skip : it
     run(c.c.name, async () => {
       const engineWasm = createWasmOnigLibWrapper()
-      const engineJs = createJavaScriptRegexEngine()
+      const engineJs = createJavaScriptRegexEngine({
+        forgiving: true,
+        target: 'ES2024',
+      })
 
       const shiki1 = await createHighlighterCore({
         langs: c.lang,


slevithan:
@antfu Would it be possible for you to modify or take over this PR to fix the tests?

- GitHub Actions / test (18.x, ubuntu-latest) has multiple failures with "SyntaxError: Invalid flags supplied to RegExp constructor 'dgv'".
  - This is expected because flag `v` requires Node.js 20+. Maybe the tests could be made to run with Oniguruma-To-ES's `target` set to `'ES2018'` instead of `'ES2024'`. That would allow it to work on Node.js 10+.
  - IMO it will be best to have the JS engine report still run with `target: 'ES2024'`.
- `packages/engine-javascript/test/compare.test.js` has multiple failures, but I think these tests need to be regenerated or deleted (not sure which).

packages/core/rollup.config.mjs
```diff
@@ -36,11 +36,9 @@ const external = [
   'hast',
   '@shikijs/vscode-textmate',
 
-  // Externalize them to make it easier to patch and experiments
+  // Externalize to make it easier to patch and experiment
   // Versions are pinned to avoid regressions
-  // Later we might consider to bundle them.
-  'oniguruma-to-js',
-  'regex',
+  'oniguruma-to-es',
```
slevithan, line 41: Not sure whether `oniguruma-to-es`'s dependencies should also be externalized. I removed the comment about considering bundling because I'd much prefer not to do that if it means these projects will lose much of their download stats (which helps add to their credibility which is really valuable for now, especially while they're new projects).

antfu:
What do you suggest for the better name of `simulation`?

packages/core/rollup.config.mjs
antfu: Totally! If `oniguruma-to-es` gets externalized, then its dependencies will be as well.

slevithan:
> What do you suggest for the better name of `simulation`?

Hm, I think the option should be removed entirely.

The last remaining thing it's doing when the option is on is `p = p.replaceAll('(^|\\\uFFFF)', '(^|\\G)')`, with a comment that this is a workaround for markdown grammars with vscode-textmate handling. I left this in place because I didn't understand the issue with vscode-textmate.

But after looking into this a big more, I think it should just be removed. Running this line doesn't change the results. vscode-textmate does weird shit with anchors `\A` and `\G` that is hard to follow or understand (presumably to allow these anchors to work while doing line-based rather than string-based processing). But Oniguruma-To-ES always converts these anchors to something else as part of emulation, so as far as I understand, vscode-textmate will never see these anchors anyway coming from Shiki's JS engine. I think this was also already true with oniguruma-to-js.

Aside: vscode-textmate also does weird shit with numbered backreferences in `end` patterns, since TextMate grammars allow using backreferences in `end` patterns that refer to captures in `begin` patterns (even though backreferences to undefined groups are errors in both Oniguruma and JS with flag `u`/`v`). This wasn't a problem with oniguruma-to-js because it used JS's RegExp legacy mode without flag `u` or `v`, so using e.g. `\1` in a regex when there were no capturing groups defined didn't error, and the `\1` was treated as an octal by JS (equivalent to `\x01`), and then it was replaced by vscode-textmate with the substring matched by capture 1 from the `begin` pattern. But since Oniguruma-To-ES doesn't use JS's legacy Unicode-unaware mode (thereby improving a variety of things), this was a problem since, initially, many `end` patterns were erroring. I solved this with what I think was a [clever hack](https://github.com/slevithan/oniguruma-to-es/blob/449aba485d494c0ce54acefc6d6f01661a13fd30/src/transform.js#L541-L554), which is the reason Oniguruma-To-ES's `tmGrammar` option exists.

> Interestingly the regex behave different on macOS and Windows/Linux 😇

What!?? In which JS environment? With Safari? If you can identify specific regexes that are behaving differently, maybe I can identify JS engine bugs and work around them.

slevithan:
> What do you suggest for the better name of `simulation`?

What about just removing the option? When if ever do you need to run with `simulation: false`?

PS: vscode-textmate's special/hacky handling of `\G` anchors is definitely causing problems that result in more mismatches (see https://github.com/shikijs/shiki/issues/828#issuecomment-2476364529).

Aside: vscode-textmate also does special/hacky things with numbered backreferences in `end` patterns, since TextMate grammars allow using backreferences in `end` patterns that refer to captures in `begin` patterns (even though backreferences to undefined groups are errors in both Oniguruma and JS with flag `u`/`v`). This wasn't a problem with the current engine because oniguruma-to-js used JS's RegExp legacy mode without flag `u` or `v`, so using e.g. `\1` in a regex when there were no capturing groups defined didn't error, and the `\1` was treated as an octal escape by JS (equivalent to `\x01`), and then it was replaced by vscode-textmate with the substring matched by capture 1 from the `begin` pattern. But since Oniguruma-To-ES doesn't use JS's legacy Unicode-unaware mode (thereby improving a variety of things), this was initially a problem (during earlier development) since many `end` patterns were erroring. I solved this with what I think was a [clever hack](https://github.com/slevithan/oniguruma-to-es/blob/449aba485d494c0ce54acefc6d6f01661a13fd30/src/transform.js#L541-L554), which is the reason Oniguruma-To-ES's `tmGrammar` option exists.

> Interestingly the regex behave different on macOS and Windows/Linux 😇

What?? 😖 In which JS environment? With Safari? I don't have access to macOS at the moment, but if you can identify specific regexes that are behaving differently, maybe I can identify JS engine bugs and work around them.

antfu:
> provide a different JS regex constructor

I think it would still be valuable, people could copy our implementation of using `Oniguruma-To-ES` and make some changes like the options, or do a preprocess/postprocess, or have a legacy implementation using `oniguruma-to-js` for legacy env that doesn't support `v` flag, etc.

> In which JS environment? 

Node.js, see the CI. I am using macOS, and I updated the snapshot in a commit, which result the snapshot mismatch on Linux and Windows. If you update it, it would be another way around. I haven't dig into which regex is causing that yet tho

slevithan:
> I think it would still be valuable, people could copy our implementation of using `Oniguruma-To-ES` and make some changes like the options, or do a preprocess/postprocess, or have a legacy implementation using `oniguruma-to-js`

That's all reasonable. I won't push on this.

> for legacy env that doesn't support `v` flag

IMO this is the strongest use case, but we can make Shiki's JS engine support this out of the box by accepting a `target` option, with the default being `'ES2024'`, and then providing that to `Oniguruma-To-ES`. If people set `target` to `ES2018`, `Oniguruma-To-ES` would use flag `u` which is supported super broadly (only a handful of grammars would lose support).

> Node.js, see the CI. I am using macOS, and I updated the snapshot in a commit, which result the snapshot mismatch on Linux and Windows. If you update it, it would be another way around. I haven't dig into which regex is causing that yet tho

Whoa. That's really tricky if the regex engine in the same version of Node.js is behaving differently on macOS and Linux/Windows.

But yeah, if you're able to identify particular regexes causing it, I'll do everything I can to try to work around it.

slevithan:
BTW, I've got a breakthrough coming in the next version (in the works). The number of supported grammars will go way up. 😊

packages/engine-javascript/src/index.ts
```diff
@@ -16,12 +16,27 @@ export interface JavaScriptRegexEngineOptions {
...
+   * Which requires Node.js 20+ or Chrome 112+.
+   * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/unicodeSets
+   *
+   * For the maximum compatibility, you can set it to `ES2018`. Which will use the `u` flag to simulate and will be less accurate.
+   *
+   * Set to `auto` to detect the target version automatically.
+   *
+   * @default 'auto'
+   */
+  target?: 'ES2024' | 'ES2025' | 'ES2018' | 'auto'
```
antfu, line 38: I implemented a `auto` target here, do you think it's worth to bringing it to `oniguruma-to-es`?

slevithan: Yes; great idea. And after I move `target: 'auto'` into `oniguruma-to-es`, it will also be able to use target `'ES2025'` when supported, making results even better.

slevithan:
Follow up PRs to further improve grammar support, etc. in #836, #842, #847, #850.

Issue #792, opened 2024-09-27
`options.meta` does not return parsed metadata object in transformer


nikolailehbrink:
### Validations

- [X] Follow our [Code of Conduct](https://github.com/shikijs/shiki/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://github.com/shikijs/shiki/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

### Describe the bug

I am trying to create a Shiki transformer that retrieves metadata like the filename from the `options.meta` object during the preprocessing step, but I'm encountering an issue.

According to the [Shiki documentation on transformers](https://shiki.matsu.io/guide/transformers#meta), the `options.meta` object should return a parsed key-value structure like:

```json
{
  "meta": "here",
  "__raw": "meta=here"
}
```

However, when I use `@shikijs/rehype` and attempt to access `options.meta`, I only receive the `__raw` string without the parsed key-value pairs. For example:

````md
```tsx twoslash filename=app/routes/test.ts
console.log("Hello, World!");
```
````

I expect `options.meta` to return:

````json
{
  "twoslash": "something",
  "filename": "app/routes/test.ts",
  "__raw": "twoslash filename=app/routes/test.ts"
}
````

But instead, I get:

```json
{
  "__raw": "twoslash filename=app/routes/test.ts"
}
```

This forces me to manually parse the `__raw` string, which contradicts the documentation.

**Steps to Reproduce:**
1. Use `@shikijs/rehype` with the following transformer:
   
   ```ts
   const transformer = () => ({
     name: "transformer",
     preprocess(code, options) {
       console.log(options.meta);
     }
   });
   ```

2. Process a code block with metadata (e.g., `twoslash filename=app/routes/test.ts`).

**Expected Behavior:**
`options.meta` should return a parsed object with metadata fields like `filename`, along with the `__raw` string.

**Actual Behavior:**
Only the `__raw` string is returned, without parsed metadata.

**Environment:**
- Vite configuration with `@shikijs/rehype` and transformers (see below).

```ts
import { defineConfig } from "vite";
import tsconfigPaths from "vite-tsconfig-paths";
import mdx from "@mdx-js/rollup";
import rehypeShiki from "@shikijs/rehype";

const transformer = () => ({
  name: "transformer",
  preprocess(code, options) {
    console.log(options.meta);
  },
});

export default defineConfig({
  plugins: [
[...]

PR #834, opened 2024-11-13
docs(transformers): clarify that shiki does not parse meta

olets:
### Description

Currently the docs imply that Shiki provides a parsed meta object. This change removes that implication. This change also provides an example of dashcased meta attribute keys and quoted meta attribute values.

### Linked Issues

- #792


cworld1:
If you want to help the reader more, I think it is very appropriate to provide a link to the issue #792 here, at least to let the reader know that there is a solution that may be helpful.

Issue #770, opened 2024-09-05
Using transformerNotationHighlight in JSX code


Yonom:
### Validations

- [X] Follow our [Code of Conduct](https://github.com/shikijs/shiki/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://github.com/shikijs/shiki/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

### Describe the bug

````ts
```ts
const Test: React.FC = () => {
  return (
    <div>
      <div /> {/* [!code highlight] */}
    </div>
  );
};
```
````

gets converted to

```ts
const Test: React.FC = () => {
  return (
    <div>
      <div /> {}
    </div>
  );
};
```

I expected no empty brackets `{}` to remain

### Reproduction

-

### Contributes

- [ ] I am willing to submit a PR to fix this issue
- [ ] I am willing to submit a PR with failing tests

Yonom:
I believe the regex here would have to be adjusted to also capture the curly brackets {}

https://github.com/shikijs/shiki/blob/eae73919d4a9c65722acb7455668e3a8f1c31fac/packages/transformers/src/transformers/notation-map.ts#L27

fuma-nama:
````md
```tsx
const Test: React.FC = () => {
  return (
    <div>
      <!-- [!code highlight] -->
      <div />
    </div>
  );
};
```
````

**Comment by phonzammi** [comment]
Based on PR #835 ([JSX section](https://github.com/shikijs/shiki/pull/835#:~:text=JSX,%3C/div%3E)) and https://github.com/shikijs/shiki/pull/835#discussion_r1841733450, it seems we can place the highlight/diff comment above the line we want to highlight.

~~~mdx
```tsx
const Test: React.FC = () => {
  return (
    <div>
      {/* [!code highlight] */}
      <div />
    </div>
  );
};
```
~~~

PR #835, opened 2024-11-14
feat(transformers): introduce `matchAlgorithm` option for new matching algorithm

fuma-nama:
### Description

This PR aims to redesign notation transformer utility and its usage to improve performance and make it more intuitive.

#### Changes

- Parse comments separately and cache it into the `data` property of `code` nodes.
- Notation transformers read and replace parsed comments.
- Update original token, remove empty comment tokens and lines if necessary.

#### New Syntax

```js
console.log("hello") // [!code highlight]

console.log("highlighted") // [!code highlight:2]
console.log("highlighted")

// the comment below will be removed
// [!code highlight:2]
console.log("highlighted")
console.log("highlighted")

console.log("combine") // [!code focus] [!code highlight]
console.log("combine") // Comment Allowed [!code focus] [!code highlight]
```

#### JSX

```jsx
<div>
    <p>Hello World</p>
    {/* [!code highlight] */}
    <p>Highlighted</p>
</div>
```

#### Breaking

It only looks for the last few tokens of line to scan comments:

```jsx
console.log("// [!code highlight]"); // will not work
```

It doesn't need multiple `//` to chain notations:

```ts
console.log("test") // [!code highlight] // [!code ++]

// will cause weird result like:
console.log("test") // //
```

#### Edge cases

for usage like:

```ts
// some text [!code highlight:2]
console.log("highlighted")
console.log("highlighted")
```

line 2-3 will be highlighted, but the original comment is kept.

#### Backward Compatibility

Added a `legacy` option to force legacy behaviors:

```ts
transformerNotationFocus({ legacy: true })
transformerNotationHighlight({ legacy: true })
transformerNotationErrorLevel({ legacy: true })
```

### Linked Issues

#619 #770 

### Additional context

this PR will introduce breaking changes, I currently made a separate package to avoid the problems. Backward compatibility is possible (edit: supported), but it takes extra care and further discussion.

code implementations:
- the `createCommentNotationTransformer` parses comment with `parseComments` and cache the result.
- each parsed comment object contains related hast nodes, and `info`:
  ```
      /* hello world */
   |----|-----------|-|
   prefix   content  suffix
  ```
- remove nodes when `content` became empty.

- transformers using `createCommentNotationTransformer` still provides a regex, but the regex no longer need to match the comment


packages/transformers/src/shared/notation-transformer.ts
```diff
@@ -0,0 +1,75 @@
...
+      }
+
+      data._shiki_notation ??= parseComments(lines, ['jsx', 'tsx'].includes(this.options.lang))
+      const parsed = data._shiki_notation
+
+      for (const comment of parsed) {
+        if (comment.info[1].length === 0)
+          continue
+
+        const isLineCommentOnly = comment.line.children.length === (comment.isJsxStyle ? 3 : 1)
```
fuma-nama, line 35: it indicates the line contains only the comment itself, like:
```js
// [!code highlight]
console.log("test")
```
so we will highlight the next line and remove the original line when `content` became empty.

packages/transformers/src/transformers/notation-map.ts
```diff
@@ -24,14 +24,14 @@ export function transformerNotationMap(
 
   return createCommentNotationTransformer(
     name,
-    new RegExp(`\\s*(?://|/\\*|<!--|#|--|%{1,2}|;{1,2}|"|')\\s+\\[!code (${Object.keys(classMap).map(escapeRegExp).join('|')})(:\\d+)?\\]\\s*(?:\\*/|-->)?\\s*$`),
+    new RegExp(`\\s*\\[!code (${Object.keys(classMap).map(escapeRegExp).join('|')})(:\\d+)?\\]`),
```
fuma-nama, line 27: we no longer need to provide the regex for matching comments, it is better given that the name `createCommentNotationTransformer` should this too.

fuma-nama:
I kept the failing unit tests to experiment about some changes for backward compatibility

Fro-Q:
This is what I'm looking for! Are you still working on it ? 'cause I actually need it so bad :>

fuma-nama:
See https://github.com/fuma-nama/shiki-transformers

antfu:
Hey @fuma-nama, thanks for the awesome work and sorry for the long due. In other to process this PR, we have to first make it non-breaking and rolling out it progressively. I have push some commits to introduce the `matchAlgorithm` which will default to the old behaviour, and we will rolling it by default in v3. I also wrote some guide for people to migrate or opt-in, if I messed anything feel free to push and improve it. Thanks

PR #862, opened 2024-12-09
docs: pass CodeBlock/highlight code + lang in as props, parameter

karlhorky:
### Description

Pass in:

1. code and language via props for `<CodeBlock>` component
2. language via 2nd `highlight()` function parameter

### Linked Issues

--

### Additional context

--

### DIFF
--- a/docs/packages/next.md
+++ b/docs/packages/next.md
@@ -19,15 +19,18 @@ export default function Page() {
   return (
     <main>
       <CodeBlock lang="ts">
-        {'console.log("Hello World")'}
+        {`
+          console.log("Hello")
+          console.log("World")
+        `}
       </CodeBlock>
     </main>
   )
 }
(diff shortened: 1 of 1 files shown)


antfu:
Could you help to fix the lint? Thanks

karlhorky:
Sure thing - what do you think of the style in this commit:

- da2d7da4846dcf3706a7f67ccc42453ec1e71d6d

antfu:
It's just some demo examples; I think it's fine either way, but I would think it's better not to involve too many concepts all at once. It shouldn't be hard for developers to catch up and fix it, it's a problem unrelated to Shiki so I think it's better to not mention that. I changed it to array join to workaround that.

karlhorky:
Sounds good, thanks for the review, edit and merge!

PR #866, opened 2024-12-16
test: improve snapshot testing for js engine

slevithan:
There are still some mismatches in the JS engine I can't explain.

Beancount is a good grammar to investigate because it has a relatively small number of patterns and doesn't use any pattern features with expected potential for mismatches, like unsupported uses of `\G`.

Following are the current grammars labeled as mismatched (no parsing errors that move them to unsupported) when `oniguruma-to-es` option `allowUnhandledGAnchors` is set to `false`:

- apex
- beancount
- kotlin
- kusto

I think if we identify what's wrong with beancount we might fix all of these, plus bring down the "diff" count on some other grammars that are currently unsupported.

However, if just looking directly at the JS engine's whole results for the beancount grammar via `compare.test.js`, although it's hard for me to interpret, it looks like there might not be any pattern match differences, but rather maybe the JS engine itself is working slightly differently than the Oniguruma engine. So it might be something that requires adjusting/fixing in the JS engine code rather than in `oniguruma-to-es`.

@antfu asked me to send a draft PR to help investigate.

Note: I've removed the file system alias for `oniguruma-to-es` since I'm not sure it’s not doing anything to help. You can just change `packages/engine-javascript/package.json` to use `"oniguruma-to-es": "file:../../../oniguruma-to-es"`, during development/testing.


antfu:
Yeah, that's interesting; it's weird that it seems that in the first pass, the results are exactly the same, but somehow, the second pass has different args. I do not yet have a clue why this would happen

antfu:
Oh, I found something; the previous format of the records wasn't linear (it was grouped by regex patterns). I updated the format to be linear, and on the second match of `beancount` there is a mismatch:

CleanShot 2024-12-17 at 17 42 17@2x

You can reproduce it with the latest commit with `pnpm test verify`

slevithan:
Okay, I found the fucker. 😖 Turns out Oniguruma's `^` cannot match after a string-terminating `\n` (though it matches after every other `\n`, including the position sandwiched between `\n\n`). There is no equivalent restriction on `$`.

In other words, `oniguruma-to-es` needs to change its conversion for `^` from `/(?<=^|\n)/v` to `/(?<=^|\n(?!$))/v`.

This change will be made in the next version of `oniguruma-to-es`. Compared to the current v0.7.0, just making this change moves **beancount**, **php**, **pug**, and **rst** to the supported list, and it significantly lowers the diff count for markdown, mdc, and po. It does nothing to help apex, kotlin, or kusto (which I mentioned in my first comment for this PR).

I'd recommend landing all of these changes (perhaps after removing the beancount test from `compare.test.ts` and the files that generates, if you prefer) since they are helpful cleanup and make it easier to compare other grammars. I'll then follow up with a PR that upgrades `oniguruma-to-es` to fix `^`.

antfu (approved):

slevithan:
Heads up that the next version of `oniguruma-to-es` will take some time because, using `compare.test.ts`, I've discovered another major culprit for remaining mismatches that I want to fix before release.

It turns out that `vscode-oniguruma` enables Oniguruma option `ONIG_OPTION_CAPTURE_GROUP` whenever options aren't manually provided (!) (see [here](https://github.com/microsoft/vscode-oniguruma/blob/1970c417eb0ebcaf8c6607774934ab2f89549c92/src/index.ts#L380)), which changes behavior in various ways when mixing named and unnamed capture. `oniguruma-to-es` doesn't currently support this option, and although I've now started work on it, it requires deep changes for things like subroutine and recursion handling. The good news is that adding support for this (and enabling the option in the JS engine) will dramatically reduce diff counts for remaining mismatched/unsupported grammars, and move at least one (confirmed so far) to supported.

Aside: There are also open issues here https://github.com/shikijs/shiki/issues/868 and here https://github.com/slackhq/vscode-hack/pull/162 that, if resolved, would further improve the JS engine's robustness.

PR #872, opened 2024-12-21
feat(engine-js): update deps, support new languages

slevithan:
Bumps `oniguruma-to-es` to [0.8.1](https://github.com/slevithan/oniguruma-to-es/releases), moving Markdown and PO from Mismatched to Supported in the JS engine report.


antfu (approved):
Getting Markdown to work is a huge milestone! Thanks for the hard work!

PR #874, opened 2024-12-21
Use correct method in docs example

alexguja:
Update import in the example docs

### DIFF
--- a/docs/packages/next.md
+++ b/docs/packages/next.md
@@ -54,7 +54,7 @@ import type { BundledLanguage } from 'shiki'
 import { toJsxRuntime } from 'hast-util-to-jsx-runtime'
 import { Fragment } from 'react'
 import { jsx, jsxs } from 'react/jsx-runtime'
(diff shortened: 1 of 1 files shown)


docs/packages/next.md
```diff
@@ -54,7 +54,7 @@ import type { BundledLanguage } from 'shiki'
 import { toJsxRuntime } from 'hast-util-to-jsx-runtime'
 import { Fragment } from 'react'
 import { jsx, jsxs } from 'react/jsx-runtime'
-import { codeToHast } from 'shiki'
+import { codeToHtml } from 'shiki'
```
alexguja, line 57: it could also be that L78 meant to be 

```tsx
async function CodeBlock(props: Props) {
  const out = await codeToHast(props.children, {
...
```

antfu: Yeah it should be another way around

alexguja: fixed 👌🏻

PR #880, opened 2024-12-31
feat: introduce precompiled-grammars

### LINKED ISSUE #878: Precomputing JS regexes for perf
Do you think it would be a good idea to precompute regexes used by the JS engine so that the regex transpilation step isn't needed at runtime? Perhaps this would be optional. Presumably, it would cause some challenges for the fine-grained bundle approach, but maybe that could be avoided/simplified in a new major Shiki version.

With some extra, somewhat-complex work in the build script for grammars, this could also make the JS regexes used by some grammars faster, by avoiding the use of a `RegExp` subclass when it's not needed (details follow).

---

`oniguruma-to-es`'s `toRegExp` outputs a `RegExp`, which might be a regexp literal or might be an instance of its `EmulatedRegExp` subclass. The subclass has some minor perf overhead, which presumably can add up for large grammars applied to large source files.

`oniguruma-to-es` uses `EmulatedRegExp` for two kinds of cases:

1. Regexes that wouldn't be emulatable otherwise. Currently this applies only to regexes that use `(^|\G)`, `(?!\G)`, or similar.
2. Regexes that would match the same strings without the subclass, but might then have differences compared to Oniguruma for subpattern details on the result array (visible from outside the regex on backreference values by index, `groups`, `indices`, and `indices.groups`). This applies when using atomic groups, possessive quantifiers, subroutines, and some uses of recursion. In some cases the subclass simply hides injected "emulation groups" from results, and in other cases it's more sophisticated, transferring subpattern match values/indices in order to accurately emulate complex edge-case effects of Oniguruma subroutines.

`oniguruma-to-es` includes an option (`avoidSubclass`) that avoids the use of a subclass (and any changes to the pattern that would rely on the subclass to be interpreted correctly). This is relevant to perf because, although enabling `avoidSubclass` for regexes in case 1 (above) results in regexes not being emulatable (and throwing), for case-2 regexes it results in faster regexes that still match the same strings.

The trick is that, if a particular regex in a grammar doesn't reference subpattern matches from outside the regex to do any additional handling, you can get a safe perf increase by not using the subclass. The build step that precomputes regexes could determine this based on whether the grammar refers to subpatterns for a given regex (using properties `captures`, `beginCaptures`, etc. in the JSON). If it doesn't, you could either:

- Check whether `toRegExp` throws with `avoidSubclass`, or...
- Use `toDetails` instead of `toRegExp` and check whether the result includes [`.subclass.strategy`](https://github.com/slevithan/oniguruma-to-es/blob/7df3c55ae6fbc24b5c551bb3e0294034160213f8/src/index.js#L77). If so, `EmulatedRegExp` needs to be used as a constructor to handle case 1, but otherwise you can run it again with `avoidSubclass` and then just use a regexp literal or `RegExp` as the constructor.

This extra handling when precomputing regexes would be totally optional, but should provide a perf increase for some grammars.

Perhaps this perf optimization could even be built into the current JS engine, if the `patterns` passed to its `JavaScriptScanner` could include data about whether the patterns will have their subpatterns referred to from outside the regexes.

---

Feel free to close this if you think it's not the right road to go down for whatever reason.

Aside: Just let me know @antfu if you're ever interested in being added as a collaborator for `oniguruma-to-es`. I'd be happy for you to make even fundamental changes that you think would be helpful.

antfu:
resolves #878 

Relies on:
- #879 
- https://github.com/shikijs/vscode-textmate/pull/3
- https://github.com/slevithan/oniguruma-to-es/pull/15

This PR introduces a new `@shikijs/langs-precompiled` package that inlines the compiled RegExp literals inside the grammar objects. And a new `@shikijs/engine-javascript/raw` that only takes precompiled grammar to further optimize the performance and bundle size


slevithan:
`oniguruma-to-es` [v0.10.0](https://github.com/slevithan/oniguruma-to-es/releases) has been published. It includes a `rawArgs` property on `EmulatedRegExp` instances so they can be accurately serialized.

`rawArgs` type:

```
{
  pattern: string;
  flags: string;
  options: {
    strategy?: string;
    useEmulationGroups?: boolean;
  };
}
```

slevithan:
Although you're not currently using `toDetails`, heads up that v0.10.0 also renamed `toDetails`'s `subclass` property as `options`, for consistency.

Also, if you bump to v0.10.0 as part of this PR, it's worth subsequently bumping `tm-grammars` and running `report-engine-js`. Although v0.10.0's improvements don't change support for existing grammars, a couple of grammars have now landed upstream fixes.

antfu:
CleanShot 2025-01-02 at 14 00 41@2x

Did a rough benchmarking. The performance still varies with different languages, but it's indeed faster than the pure JS version.

packages/langs-precompiled/scripts/langs.ts
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  // Boolean or number
+  if (typeof value === 'boolean' || typeof value === 'number') {
+    return String(value)
+  }
+
+  if (value instanceof EmulatedRegExp) {
+    return `new EmulatedRegExp(${JSON.stringify(value.rawArgs.pattern)},${JSON.stringify(value.rawArgs.flags)},${JSON.stringify(value.rawArgs.options)})`
+  }
```
slevithan, line 123: If this code could somehow know whether the regex it's evaluating is used with captures for the grammar, it could get a perf boost using the trick described in #878 by avoiding the use `EmulatedRegExp` when capture values are not used.

```js
  if (value instanceof EmulatedRegExp) {
    if (regexUsedWithCaptures && !value.rawArgs.options.strategy) {
      return `new EmulatedRegExp(${JSON.stringify(value.rawArgs.pattern)},${JSON.stringify(value.rawArgs.flags)},${JSON.stringify(value.rawArgs.options)})`
    }
    // If emulation `strategy` not used and we don't rely on subpattern values/positions, can just use a regex literal
    return value.toString()
  }
```

antfu: Would you like to send a PR for that? Thanks!

PR #909, opened 2025-01-27
fix(transformers): fix matching indices for word-highlight

### LINKED ISSUE #908: `transformerMetaWordHighlight` fail to highlight overlapping text instances
### Validations

### Describe the bug

When using `transformerMetaWordHighlight` to highlight words that overlap, Shiki fails.

Example:

```
```js /xx/
xxx
``` .
```

Shiki will try to create two `spans` with class `highlighted-word`: `(xx)x` and `x(xx)` and will result in the following error:

```
Failed to parse Markdown file "...":
Decorations {"line":0,"character":0,"offset":0} and {"line":0,"character":1,"offset":1} intersect.
```

Perhaps a way to fix is to allow highlighting words by character position?

On a related note, why the use of `/.../` that resembles regex?

### Reproduction

https://stackblitz.com/edit/shiki-intersect?file=src%2Fpages%2Findex.astro,astro.config.mjs,src%2Fcomponents%2Ftest.md

### Contributes

artt:
### Description

Partially fixes #908 and add tests.

### Additional context

Users still cannot freely choose which instance the highlight should be on (regex will choose whatever comes up first).

It would be nice to allow users to manually specify which character (by index) the highlight should be. For example, `js {1[6:12],3-4} to highlight characters 6 through 12 of the first line, and lines 3+4. But that seems like a lot more work. At least this PR will eliminate the error.


packages/transformers/src/transformers/meta-highlight-word.ts
```diff
@@ -55,11 +55,13 @@ export function transformerMetaWordHighlight(
...
 }
 
-function findAllSubstringIndexes(str: string, substr: string): number[] {
+export function findAllSubstringIndexes(str: string, substr: string): number[] {
   const indexes = []
-  let i = -1
-  // eslint-disable-next-line no-cond-assign
-  while ((i = str.indexOf(substr, i + 1)) !== -1)
-    indexes.push(i)
+  const re = new RegExp(substr, 'g')
```
antfu, line 60: I guess we need to properly escape the `substr` if we want to use that approach.

artt: Not sure if this is already handled by this function `parseMetaHighlightWords`. If you could help give some examples that would be awesome.

https://github.com/shikijs/shiki/blob/0b9137f8413eb1f276928707c0ac92de0664a314/packages/transformers/src/transformers/meta-highlight-word.ts#L3C1-L13C2

And, do we want to use this approach?

slevithan: The safest way to escape regex special characters is to use [Regex+](https://github.com/slevithan/regex) and interpolate the string to escape. Nearly as safe but with noisier output would be to use native `RegExp.escape` (an ES proposal), which isn't supported by browsers yet. The most popular way to escape regex special characters (based on npm download stats) is to use the unsafe but lightweight [escape-string-regexp](https://github.com/sindresorhus/escape-string-regexp). Context safety isn't relevant when the entire regex pattern is an escaped string, though, so no worries.

If it was me, I might just add an inline `.replace(/[|\\{}()[\]^$+*?.]/g, '\\$&')` with no libraries. This doesn't escape chars that might need to be escaped based on context (`-`, `,`, digits, etc.), since no context awareness is needed when the escaped string is being used as the entire regex pattern.

But then, I don't understand why you're moving from string search to regex search in the first place.

`parseMetaHighlightWords` isn't escaping special characters; it seems to be matching JS regex literals? I also notice it has a couple minor issue issues with that: it will inappropriately match `/foo\/` (no unescaped trailing `/`) and it will not correctly match `/[/]/` (unescaped `/` is allowed in JS character classes, unless the regex uses flag `v`).

packages/transformers/src/transformers/meta-highlight-word.ts
```diff
@@ -55,11 +55,13 @@ export function transformerMetaWordHighlight(
...
+  let match = re.exec(str)
+  while (match !== null) {
```
slevithan, line 62: This looks like it will loop forever if `re` matches an empty string.

packages/transformers/src/transformers/meta-highlight-word.ts
artt: > But then, I don't understand why you're moving from string search to regex search in the first place.

Problem from #908. But yes, this is one out of the many ways to solve it. And it doesn't solve all the problems. Maybe I'll close this first and move it to discussion to see what people think first.

antfu:
I see the problem. It runs out a bug, and we don't need regex for that

PR #922, opened 2025-02-07
fix: node10 types and implementation resolution

userquin:
### Description

This PR includes the following fixes for node10 module resolution:
- add missing types
- fix wrong resolution and implementation in VP: `@shikijs/vitepress-twoslash/cache-fs` resolving to default package export (expand https://arethetypeswrong.github.io/?p=%40shikijs%2Fvitepress-twoslash%402.3.2 then expant entrypoints/node10 resolution and implementation and check the filename, check screenshots below)

/cc @antfu 

About the *.mts  `typesVersions`  mapping in `langs`, `langs-precompiled` and `theme` packages, it is a hack to allow resolve both types and implementation correctly, I've send you a few DM at discord with some screenshots with the traces for both.

I also need to review a few repositories :sweat_smile: 

### Additional context

image
_current cache-fs subpackage export_

image
_cache-fs subpackage export with this PR_

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -31,6 +31,13 @@
   "main": "./dist/index.mjs",
   "module": "./dist/index.mjs",
   "types": "./dist/index.d.mts",
+  "typesVersions": {
+    "*": {
+      "cli": [
+        "./dist/cli.mts"
+      ]
+    }
+  },
   "bin": {
     "shiki": "bin.mjs",
     "shiki-cli": "bin.mjs",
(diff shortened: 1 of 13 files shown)


userquin:
I need to review why node10 resolving implementation to d.mts instead .mjs

antfu:
Can we drop Node 10 already? 🤣

userquin:
> Can we drop Node 10 already? 🤣

Yeah, it is a pain... in fact your blog about [Types for Submodules](https://antfu.me/posts/types-for-sub-modules) is wrong (`@tmkx shared me`), you shouldn't use the dts in typesVersions, the implementation resolution will resolve to the dts instead the implmentation (here to mjs files), you need to use `.ts` (if using compatible) or `.mts` (if using ESM only).

PR #955, opened 2025-03-01
docs(markdown-it): work around fence rule pre code wrapper, close #811

### LINKED ISSUE #811: using a transformer to add a wrapping element results in extra duplicative nodes
### Validations

### Describe the bug

I'm working on a transformer which adds a root wrapper. Minimal goal is

```html
<div data-my-attr="val">
  <pre class="shiki <theme>" style="…" tabindex="0">
    <code class="language-js">
    …
    </code>
  </pre>
</div>
```

I'm using `@shikijs/markdown-it`.

### root hook

My first thought was to use the `root` hook:

```js
        root(hast) {
          return [
            {
              type: 'element',
              tagName: 'div',
              properties: {
                'data-my-attr': 'val',
              },
              children: hast.children,
            },
          ];
        },
```

Given

````md
```js
```
````

that transformer results in

```html
<pre> <!-- unexpected wrapper 1 -->
  <code class="language-js"> <!-- unexpected wrapper 2 -->
    <div data-my-attr="val">
      <pre class="shiki <theme>" style="…" tabindex="0">
        <code class="language-js">
          <span class="line"></span>
        </code>
      </pre>
    </div>
  </code>
</pre>
```

### preprocess and postprocess hooks

Using a postprocess hook is a better fit for my need, because I'd like to use `options.meta`.

That doesn't work either:

```js
        postprocess(html, options) {
          return `<div data-my-attr="val">${html}</div>`;
        },
```

Given the same markdown as in the `root` hook example above, I get the same problematic result as above.

### Reproduction

.

### Contributes

olets:
### Description

`markdown-it` wraps code block html in `<pre><code>`.  That can be undesirable when using a Shiki transformer which adds its own `<pre><code>`. This PR

- documents this fact
- points to the `markdown-it` discussion which is the closest thing to official documentation of how to work around this
- and links to the package I put together based on that following discussion in a Shiki issue

### Linked Issues

- Closes #811

### Additional context

If you decide to close without merging, please also close #811.


docs/packages/markdown-it.md
```diff
@@ -97,3 +97,35 @@ md.use(
 // Use `md.renderAsync` instead of `md.render`
 const html = await md.renderAsync('# Title\n```ts\nconsole.log("Hello, World!")\n```')
 ````
+
+## Transforms
```
antfu, line 101:
```suggestion
## Transformer Caveats
```

antfu (approved):

PR #964, opened 2025-03-12
feat(vitepress-twoslash): add option for custom twoslasher

zhiyuanzmj:
### Description

Hi, I  write a [@ts-macro/twoslash](https://github.com/ts-macro/twoslash) plugin for [ts-macro](https://github.com/ts-macro/ts-macro).
So can we export an option let me to custom twoslasher?


zhiyuanzmj:
Oh, it's already has the twoslasher option in twoslash.

PR #976, opened 2025-03-25
feat: add handling for `FontStyle.Strikethrough`

### LINKED ISSUE #975: Add support for strikethrough in ANSI
### Validations

### Describe the bug

Currently the ANSI support for font-styles does not handle strikethrough, although that font-style is supported by the library that Shiki depends on.

https://github.com/shikijs/shiki/blob/f8ef446af3f5f3910131b692e0c81ce032754ffa/packages/core/src/highlight/code-to-tokens-ansi.ts#L50-L58

Support for strikethrough would make Shiki closer to complete ANSI support.

### Reproduction

Reproduction not needed as affected code block is referenced in description.

### Contributes

- [x] I am willing to submit a PR with failing tests

dhruvkb:
### Description

This PR adds support for handling the `FontStyle.Strikethrough` in a few places where it was not present. The primary motivation was to support ANSI strikethrough which is supported by the underlying `ansi-sequence-parser` package but not by Shiki.

### Linked Issues

Fixes #975 by myself

### Additional context

I would appreciate any help in pointing out the new test cases which should be added to cover the changed code (and where to add them).

### DIFF
--- a/docs/.vitepress/theme/transformers.css
+++ b/docs/.vitepress/theme/transformers.css
@@ -23,16 +23,24 @@
   border-radius: 4px;
 }
 
+.vp-doc [class*='language-'] pre {
+  background-color: var(--shiki-light-bg, inherit);
+}
+
+.dark .vp-doc [class*='language-'] pre {
+  background-color: var(--shiki-dark-bg, inherit);
+}
+
 .vp-code,
-.vp-code span {
-  /* background-color: var(--shiki-light-bg, inherit); */
+.vp-code span:not(.line.highlighted):not(.line.highlighted *) {
+  background-color: var(--shiki-light-bg, inherit);
   text-decoration: var(--shiki-light-text-decoration, inherit);
   font-weight: var(--shiki-light-font-weight, inherit);
 }
 
 .dark .vp-code,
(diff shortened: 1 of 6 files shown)


docs/.vitepress/theme/transformers.css
```diff
@@ -22,3 +22,17 @@
   margin: -1px -3px;
   border-radius: 4px;
 }
+
+.vp-code,
+.vp-code span {
+  /* background-color: var(--shiki-bg, inherit); */
```
dhruvkb, line 28: When this is enabled, ANSI background colors become visible but this breaks all other code examples in the docs.

image

image

That's why I've set this disabled currently, but then ANSI background colors don't render at all (and consequently the reversed style).

image

dhruvkb (commented):
It was actually pointed out to me by @hippotastic in expressive-code/expressive-code#316.

docs/.vitepress/theme/transformers.css
dhruvkb: To work around this, had to write a bunch of inelegant CSS in 1a6bb6f. @antfu if you have any advice on what can be done about that, please do share!

PR #991, opened 2025-04-15
feat(core): add option to merge consecutive tokens with same style (#972)

### LINKED ISSUE #972: Feature proposal: concat spans with the same style

Here is a snippet of Shiki output (yaml, `min-light`):

```html
<span class="line">
  <span style="color:#D32F2F;">  image</span>
  <span style="color:#D32F2F;">:</span>
  <span style="color:#22863A;"> node:20-alpine</span>
</span>
```

As you can see, `  image` and `:` spans have the same color. It doesn't make sense to render them as a separate `<span>`s, and a single one `  image:` would be sufficient.

### Suggested solution

Concatenate consecutive spans with the same styling:

```html
<span class="line">
  <span style="color:#D32F2F;">  image:</span>
  <span style="color:#22863A;"> node:20-alpine</span>
</span>
```

### Alternative

_No response_

### Additional context

This change could significantly reduce amount of generated HTML and DOM nodes, which is clearly a performance benefit in web.

The gain varies depending on a theme. The less variety of colors there is, the more compaction effect there will be.

### Validations

- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

yvbopeng:
### Description

This PR addresses issue #972 by introducing a new option `mergeSameStyleTokens` in `CodeToHastOptions`. This option allows users to merge consecutive tokens with identical styles, reducing the number of DOM nodes and improving rendering performance.

### Linked Issues

fixes #972 

### Additional context

- Added `mergeSameStyleTokens` option to `CodeToHastOptions` with a default value of `false`.
- Implemented `mergeAdjacentStyledTokens` function to merge tokens with the same style.
- Updated `codeToHast` to apply token merging based on the new option.
- Documented the new option in the code comments.
- The new option is disabled by default, ensuring no breaking changes.
- When enabled, it optimizes the generated HTML by reducing `<span>` elements, which can significantly improve performance for large code blocks.


antfu:
Can you add some snapshot tests for it to gurad the behaviour? Thanks

antfu (approved):

PR #997, opened 2025-04-22
fix(colorized-brackets): handle correctly with dual themes

### LINKED ISSUE #989: Adjacent brackets are colored incorrectly on https://shiki.style/packages/colorized-brackets
### Validations

### Describe the bug

Colorized brackets are not working correctly on https://shiki.style/packages/colorized-brackets. Brackets that are right next to each other all share a color. See for this screenshot:

Image

There are many instances of the bug in the screenshot, for example in the second to last line, `})]` should be 3 different colors, but they are all the same color.

I haven't investigated the code yet, but a few observations:
- it looks like each "group" of brackets all gets the color of the final bracket in the group
- other punctuation seems to be affected too; see some of the commas and colons in the screenshot
- apparently colorized bracket tests are still passing, so this looks to be specific to some config
- this documentation page was colored correctly when colorized-brackets package was first introduced

### Reproduction

Run the shiki docs site, go to the /packages/colorized-brackets.

### Contributes

oatmealproblem:
### Description

This fixes a few issues with the @shikijs/colorized-brackets plugin:
- brackets colors could be incorrect when using dual themes (caused by shared `htmlStyle` objects)
- bracket colors didn't match theme when using Expressive Code (caused by EC appending IDs to theme names)
- some minor typos

### Linked Issues

- Fixes #989 (caused by the first issue)


packages/colorized-brackets/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@shikijs/colorized-brackets",
   "type": "module",
   "version": "3.2.2",
-  "description": "Collective of common transformers transformers for Shiki",
+  "description": "VSCode-style colorized brackets transformer for Shiki",
```
antfu, line 5: Is this change related?

antfu (approved):

Issue #710, opened 2024-06-24
`@shikijs/rehype`: support `langAlias` options


nomyfan:
### Clear and concise description of the problem

`createHighlighter` supports `langAlias`. It would be nice if rehype supports it too.

### Suggested solution

Add `langAlias` option for rehype plugin and pass down to `createHighlighter`

### Validations

- [X] Follow our [Code of Conduct](https://github.com/eslint-stylistic/eslint-stylistic/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://eslint.style/contribute/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Contributes

- [X] If this feature request is accepted, I am willing to submit a PR to fix this issue

PR #1015, opened 2025-05-12
feat(rehype): add langAlias option

SoonIter:
### Description

feat(rehype): add langAlias option

### Linked Issues

close https://github.com/shikijs/shiki/issues/710


SoonIter:
https://github.com/SoonIter/shiki/blob/391b095b1d796877186f9fccd411a09eaf24e5f0/packages/core/src/constructors/bundle-factory.ts#L183

The singleton pattern caused the test case in rehype to fail. Running multiple test cases is error, but running a single case is passed.

IMO, rehype-shiki should be stateless, I will temporarily pass by bringing my case greenify the CI.

PR #1025, opened 2025-05-22
chore(license): update copyright info in LICENSE file

1411430556:
### Description

**_update copyright info in LICENSE file_**


antfu:
Thank you but that's not how it works, we don't update it every year.

PR #1050, opened 2025-08-03
Create devcontainer.json

brittanyshelton23:


antfu:
Thank you but we are not interested in adding that.

Issue #906, opened 2025-01-24
Display funding/sponsor links for grammars and themes?


jtbandes:
Hi there,
Since Shiki uses grammars and themes produced by many open source authors, I was wondering if you would be interested in displaying funding/sponsorship links directly in the docs or website somewhere?

Locally I developed a POC to modify [github.ts](https://github.com/shikijs/textmate-grammars-themes/blob/main/scripts/shared/github.ts) to fetch the `FUNDING.yml` files for each referenced repo. What I found is:
- 17 of the `tm-themes` have a FUNDING.yml
- 37 of the `tm-grammars` have a FUNDING.yml

(the number of _unique_ ones is a bit smaller because some people are authors of a few different themes/grammars).

I think we could consider displaying these "❤️" links:
- [x] On https://shiki.style/themes & https://shiki.style/languages (a new column in the table?) https://github.com/shikijs/shiki/pull/1052
- [x] In the [tm-grammars readme](https://github.com/shikijs/textmate-grammars-themes/tree/main/packages/tm-grammars) & [tm-themes readme](https://github.com/shikijs/textmate-grammars-themes/tree/main/packages/tm-themes) (also new column?) https://github.com/shikijs/textmate-grammars-themes/pull/116
- [x] Maybe in the grammar/theme playground app as well: https://github.com/shikijs/textmate-grammars-themes/pull/150
- open to any different suggestions...

(It would also need to parse the funding info according to [docs](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/displaying-a-sponsor-button-in-your-repository) and determine what links to display.)

If you think this is a good idea, I can clean up and submit a PR with my changes to fetch the funding info and some of the other modifications too. Looking forward to hear your thoughts!

antfu:
All sounds good! First we need to include that field in the source: https://github.com/shikijs/textmate-grammars-themes

Looking forward to that!

jtbandes:
Ready for initial feedback :) https://github.com/shikijs/textmate-grammars-themes/pull/116

slevithan:
The linked PR has now been merged. Are additional changes needed for this?

jtbandes:
Hi @antfu, just wondering if you are still interesting in displaying the funding links on shiki.style, playground, and the readmes? I know that you previously wanted to make some design changes first. Let me know if you are still interested in doing this—I could also help to implement it.

antfu:
[...]

PR #1052, opened 2025-08-04
feat: add funding buttons to themes & languages list

jtbandes:
### Description

Towards #906:
- Adds a ❤️ button to https://shiki.style/languages & https://shiki.style/themes, using funding information from https://github.com/shikijs/textmate-grammars-themes/pull/116.
- Switches the playground to import from `tm-grammars`/`tm-themes` instead of `shiki/bundle/{web,full}`. As discussed in https://github.com/shikijs/shiki/issues/906#issuecomment-3134881258, the shiki bundle does not include funding info and it's not necessary to increase the bundle size by adding it there when the same information is available from the other packages.

image

image

### Linked Issues

- #906

### DIFF
--- a/docs/.vitepress/store/playground.ts
+++ b/docs/.vitepress/store/playground.ts
@@ -1,30 +1,19 @@
 /// <reference types="vite/client" />
 
-import type { BundledLanguageInfo, BundledThemeInfo } from '@shikijs/types'
+import type { BundledLanguage, BundledTheme } from 'shiki'
+import type { GrammarInfo } from 'tm-grammars'
+import type { ThemeInfo } from 'tm-themes'
 import { useLocalStorage } from '@vueuse/core'
 import { acceptHMRUpdate, defineStore } from 'pinia'
 import { ref, shallowRef, watch } from 'vue'
 
 export const usePlayground = defineStore('playground', () => {
-  const lang = useLocalStorage('shiki-playground-lang', 'typescript')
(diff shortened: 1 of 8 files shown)


docs/.vitepress/store/playground.ts
```diff
@@ -1,30 +1,19 @@
...
-      name: 'TypeScript',
-      import: undefined!,
-    },
-  ])
-  const bundledLangsFull = shallowRef<BundledLanguageInfo[]>([])
-  const bundledLangsWeb = shallowRef<BundledLanguageInfo[]>([])
+  const lang = useLocalStorage<BundledLanguage>('shiki-playground-lang', 'typescript')
+  const theme = useLocalStorage<BundledTheme>('shiki-playground-theme', 'vitesse-dark')
+  const allThemes = shallowRef<ThemeInfo[]>([])
+  const allLanguages = shallowRef<GrammarInfo[]>([])
```
jtbandes, line 14: Is it okay to leave these empty? Since I changed the type it was harder to include some default values.

docs/package.json
```diff
@@ -16,6 +16,8 @@
     "@vueuse/core": "catalog:docs",
     "floating-vue": "catalog:docs",
     "pinia": "catalog:cli",
+    "tm-grammars": "catalog:inlined",
+    "tm-themes": "catalog:inlined",
```
jtbandes, line 20: is this catalog import ok?

PR #1060, opened 2025-08-13
feat(transformers): render indent guides

KazariEX:
### Description

Implement `editor.guides.indentation` in VSCode.

Allow using `{indent:4}` to control the indentation level of a single code block, or `{indent:false}` to turn off it.

When the indentation is on an empty line, in order for it to be positioned correctly without taking up width, I used `--indent-offset` to pass the value that needs to be assigned to the CSS `left` property. (Not sure if passing values using `--indent-size` + `--indent-level` would be more appropriate.)

### Additional context

Preview:

image


antfu:
Could you also help to update the docs? Thanks

PR #1061, opened 2025-08-17
fix(hast)!: unwrap a Root node and rename property class to className

### LINKED ISSUE #1033: Is it Appropriate for @shikijs/rehype to Output a HAST Root Node?
🖐️ Hello, I'm currently working on [`vue-markdown`](https://github.com/shunnNet/vue-markdown), which uses Unified to convert Markdown into HAST before rendering it as Vue components.

While trying to support `@shikijs/rehype`, I noticed that the HAST it produces includes a `Root` node as the outermost element, followed by the highlighted content. According to the HAST specification, the Root node should “never be a child” ([source](https://github.com/syntax-tree/hast?tab=readme-ov-file#root)). Because of this, I initially didn’t handle the Root node, which resulted in Shiki-highlighted content not being rendered correctly.

Shiki adds a Root node as a base when converting to HAST ([code reference](https://github.com/shikijs/shiki/blob/2bda4507af45662838a5405f268b2a8dc8930b3f/packages/core/src/highlight/code-to-hast.ts#L95)).

Here’s a reproduction of the issue:
https://stackblitz.com/edit/stackblitz-starters-eddwspb8?file=index.js

Since this Root node becomes a child during the Markdown rendering process, it violates the HAST specification. I’m wondering: is this output appropriate? Would you consider adding an option to avoid wrapping the result in a Root node?

tenjo-t:
### Description

- fix #1033
- rename property class to className

hast uses `className` instead of `class`. Utilities such as `hast-util-classnames` handle only `className`.

> The HTML attributes class and for respectively become className and htmlFor in alignment with the DOM. No other attributes gain different names as properties, other than a change in casing.
> <https://github.com/syntax-tree/hast?tab=readme-ov-file#propertyname>

### Linked Issues

#1033

### Additional context

`processNode` was modified in #896.

### DIFF
--- a/packages/core/src/highlight/code-to-hast.ts
+++ b/packages/core/src/highlight/code-to-hast.ts
@@ -106,7 +106,7 @@ export function tokensToHast(
     type: 'element',
     tagName: 'pre',
     properties: {
-      class: `shiki ${options.themeName || ''}`,
+      className: `shiki ${options.themeName || ''}`,
       style: options.rootStyle || `background-color:${options.bg};color:${options.fg}`,
       ...(tabindex !== false && tabindex != null)
         ? {
@@ -170,13 +170,17 @@ export function tokensToHast(
     let lineNode: Element = {
       type: 'element',
       tagName: 'span',
-      properties: { class: 'line' },
+      properties: { className: 'line' },
       children: [],
     }
 
     let col = 0
 
     for (const token of line) {
+      if (token.htmlAttrs?.class) {
+        token.htmlAttrs.className = token.htmlAttrs.class
+        delete token.htmlAttrs.class
+      }
       let tokenNode: Element = {
         type: 'element',
         tagName: 'span',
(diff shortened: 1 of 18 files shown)


antfu:
I think this would be a breaking change that we need to pending it until the next major

x-delfino:
would a patch with just the Root node unwrapping be considered for merge before next major version?

it's a single line in this PR that doesn't touch core and is fixing the non-compliance with HAST (#1033)

x-delfino:
hadn't realised the date on this PR. seems like it missed the last major release

just to confirm, I've tested the Root unwrapping on the latest release and it's working as expected. just changing:

https://github.com/shikijs/shiki/blob/2bd0948d11845c4fe6ddce5c06a6e94795af4026/packages/rehype/src/core.ts#L146

to:

```ts
        parent.children.splice(index, 1, ...fragment.children)
```

PR #1063, opened 2025-08-20
feat(vitepress-twoslash): twoslash inline type cache for markdown

serkodev:
### TODO
- [x] read / write cached twoslash types to the md source
- [x] force rebuild inline cache data
  - [ ] trigger rebuild
- [x] prune all inline cache data (todo trigger)
  - [ ] trigger prune
- [x] compress inline cache data with lz-string
- [x] make the config more easy to use

---

### Description

This PR introduce twoslash inline type cache for markdown file in vitepress during dev and build time.

- Add `// @twoslash-cache` to twoslash fence inside markdown
- Move `typesCache` option from `vitepress-twoslash` to `twoslash` package

### Additional context
Thanks @antfu for the idea.


packages/vitepress-twoslash/src/cache-inline/fence-source-map.ts
```diff
@@ -0,0 +1,162 @@
+import type { ShikiTransformer } from 'shiki'
+import type { Plugin } from 'vite'
+import type { UserConfig } from 'vitepress'
+import { readFileSync } from 'node:fs'
+import MagicString from 'magic-string'
+import MarkdownIt from 'markdown-it'
```
antfu, line 6: Is it possible for us to abstract a bit further to decouple this feature from markdown-it and Vitepress so it can be used by other tools as well in the future?

serkodev: Just did some refactoring to decouple them. Do we need to add some new exports from the package, or keep it for internal use for now?

antfu: I think we ultimately we want the core feature to be extract to the `@shikijs/twoslash` package instead of the vitepress one. So it could be used for rehype or other integrations. Let's having it for now and refactor later.

PR #1066, opened 2025-08-25
feat(vitepress-twoslash): improve UX for option customization

9romise:
### Description

This PR makes it more convenient for users to modify existing themes.

> [!IMPORTANT]
> One thing to note: I've changed the `theme` option to `themes` to align with upstream naming. This is a **breaking change**, so if you prefer, I can move this adjustment to a separate PR. Just let me know!

### Linked Issues

Related: https://github.com/vueuse/vueuse/issues/4983

### DIFF
--- a/packages/vitepress-twoslash/src/client.ts
+++ b/packages/vitepress-twoslash/src/client.ts
@@ -24,6 +24,7 @@ const TwoslashFloatingVue = {
     app.use(FloatingVue, {
       ...options,
       themes: {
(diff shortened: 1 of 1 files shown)


antfu:
Is it possible to keep backward compatibilty and mark it as deprecated?

9romise:
The option type (`FloatingVueConfig`) is `any`, and it's from upstream. Maybe I should also contribute to it for the better type.

PR #1070, opened 2025-09-01
fix: Improve Twoslash error handling

### LINKED ISSUE #1069: onTwoslashError returning string is ignored instead of replacing the code
### Validations

### Describe the bug

According to the JSDoc, if onTwoslashError returns a string, the returned string should replace the corresponding code block.
However, in the actual implementation, the returned string is ignored, and the code is not replaced.

### Expected Behavior
The returned string should replace the code block where the error occurred, as documented in the JSDoc.

### Actual Behavior
The returned string is ignored, and the original code remains unchanged.

### Reproduction

none

### Contributes

Karibash:
### Description

Fixes error handling in Twoslash packages to properly return cleaned code when errors occur, instead of ignoring the error handler's return value.

### Linked Issues

close #1069

### DIFF
--- a/packages/twoslash/src/core.ts
+++ b/packages/twoslash/src/core.ts
@@ -129,7 +129,7 @@ export function createTransformerFactory(
           catch (error) {
             const result = onTwoslashError(error, code, lang, this.options)
             if (typeof result === 'string')
-              return code
+              return result
           }
         }
       },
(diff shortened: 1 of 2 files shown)


antfu (approved):

avivkeller:
Hiya! Can this be released as a patch version? I'd love to do this feature to pass some server-side errors to the client (for debugging)

PR #1095, opened 2025-11-05
fix(lang-ansi): Fallback for missing theme ANSI colors

### LINKED ISSUE #1059: ansi highlight not work
### Validations

### Describe the bug

language `ansi` not highlight with theme one-light

### Reproduction

https://codesandbox.io/p/sandbox/8k24zd

### Contributes

Chanakyasinde:
### Description

This PR fixes an issue where ANSI language highlighting (`lang: ansi`) would fail on themes that do not provide their own `terminal.ansi*` colors (as seen with `one-light`).

The original code would attempt to read `theme.colors` for the ANSI palette. If a theme didn't define these colors, `undefined` was passed to the `createColorPalette` function, causing an error.

This change introduces a `defaultAnsiColors` map, which provides a standard VSCode-compatible set of fallback colors. The `tokenizeAnsiWithTheme` function now checks for a theme color and uses a color from this default map if one isn't found, ensuring the ANSI tokenizer always receives valid color values.

### Linked Issues

Fixes #1059

### Additional context

The `defaultAnsiColors` map is based on the standard VSCode fallbacks. This ensures consistent behavior for themes that don't specify their own terminal colors, resolving the bug while maintaining correctness for themes that do.

### DIFF
--- a/packages/core/src/highlight/code-to-tokens-ansi.ts
+++ b/packages/core/src/highlight/code-to-tokens-ansi.ts
@@ -45,8 +45,6 @@ export function tokenizeAnsiWithTheme(
     namedColors.map((name) => {
       const key = `terminal.ansi${name[0].toUpperCase()}${name.substring(1)}`
       const themeColor = theme.colors?.[key]
-
-      // ✅ Use theme color if exists, otherwise fallback
       return [name, themeColor || defaultAnsiColors[name]]
     }),
   ) as Record<string, string>
@@ -117,6 +115,7 @@ function dimColor(color: string): string {
       return `#${hexMatch[1]}${hexMatch[2]}80`
     }
     else {
+      // convert from #rgb to #rrggbb80
       return `#${Array
         .from(hexMatch[1])
         .map(x => `${x}${x}`)


packages/core/src/highlight/code-to-tokens-ansi.ts
```diff
@@ -72,29 +96,20 @@ export function tokenizeAnsiWithTheme(
...
-      // convert from #rrggbb to #rrggbb80
       return `#${hexMatch[1]}${hexMatch[2]}80`
     }
     else {
-      // convert from #rgb to #rrggbb80
-      return `#${Array
-        .from(hexMatch[1])
-        .map(x => `${x}${x}`)
-        .join('')}80`
+      return `#${Array.from(hexMatch[1]).map(x => x + x).join('')}80`
```
antfu: Why do we touch this part of the code and removing the comments?

Chanakyasinde: Hi @antfu , you're right. Those changes were unintentional, likely from my local code formatter. They aren't related to the ANSI color fallback.

I'll revert them back to its original state and push a new commit. My apologies for the noise!

Issue #1020, opened 2025-05-16
Text highlighting issue between angle brackets


itskartik22:
### Validations

- [x] Follow our [Code of Conduct](https://github.com/shikijs/shiki/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guide](https://github.com/shikijs/shiki/blob/main/CONTRIBUTING.md).
- [x] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

### Describe the bug

In bash/shell code block, when we are having some text inside angle brackets `pfinfo <usual_guide>`. A letter before the right angle bracket is showing unusual highlighting behaviour.

<img width="233" alt="Image" src="https://github.com/user-attachments/assets/4bd01163-725e-4b51-b9a9-fbdc0ac757bd" />

Please fix this asap.

### Reproduction

Reproduction

### Contributes

- [ ] I am willing to submit a PR to fix this issue
- [ ] I am willing to submit a PR with failing tests

Rudrxxx:
Hi! @antfu 
I’ve opened a PR #1147 that fixes this issue.

The problem was caused by the upstream VS Code `shellscript` grammar using a `(?!>)` negative lookahead in the `argument_context` rule, which caused the final character before `>` to fall out of the argument scope. This is why `<abcd>` highlighted `abc` correctly but dropped `d`.

My PR introduces a minimal override to correct this behavior without forking the entire grammar, and adds a regression test to ensure `<...>` arguments are fully highlighted moving forward.


Would appreciate a review when you have time!

PR #1100, opened 2025-11-06
"fix(shiki): escape <placeholders> in shell code blocks to prevent HT…

Hariksh:
…ML parsing"

### Description

This PR fixes an issue in **bash/shell code blocks** where syntax highlighting behaved incorrectly when text was enclosed within angle brackets.

**Example:**
```bash
pfinfo <usual_guide>
```

### Linked Issue

#1020

### DIFF
--- a/packages/core/src/highlight/code-to-hast.ts
+++ b/packages/core/src/highlight/code-to-hast.ts
@@ -207,6 +207,9 @@ export function tokensToHast(
       lineNodes.push(lineNode)
       lines.push(lineNode)
     }
+    else if (structure === 'inline') {
+      lineNodes.push(lineNode)
+    }
   })
 
   if (structure === 'classic') {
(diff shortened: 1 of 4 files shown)


Hariksh:
I noticed that the CI workflow is currently failing on Node **18.x** with the following error:


This appears to be caused by a recent dependency update (likely `vitest` or a related package) that depends on `node:inspector/promises`, which was only introduced in **Node 20+**.  
Since this PR only adds a preprocessing transformer (`shiki:fix-shell-angle-placeholders`) for shell code blocks in the docs, it doesn’t touch any runtime logic or test-related modules.

**Verified locally:**  
- Works fine on Node 20.x and 22.x.  
- Only Node 18.x fails due to the missing built-in module.

###  Suggested follow-up
You might consider:
- Updating the CI matrix to drop Node 18.x (since Node 18 is now EOL), or  
- Adding a version check to skip tests using `node:inspector/promises` on older Node versions.

Thanks for reviewing this!

antfu:
I am not sure if that's the correct fix. I feel that's probably something to be fixed on the upstream grammar side instead

PR #1104, opened 2025-11-07
Fix Query Popper Persistence in Twoslash Renderer

### LINKED ISSUE #981: Bug: twoslash query popper disappears when clicked
### Validations

### Describe the bug

Repoduction:

1. Go to <https://twoslash.netlify.app/refs/notations#extract-type>

Image

2. Click the popper

Image

For [extract-type](https://twoslash.netlify.app/refs/notations#extract-type) part, it is possible to click again on the `msg` identifier to make the popper reappear (and hover doesn't work!). For [completions](https://twoslash.netlify.app/refs/notations#completions), there is no way to make the popper reappear.

I think the popper generated by Queries in Twoslash should be persistent (neither responding to clicks inside the popper nor responding to hover of the trigger element).

(It would be even better if it also occupies the space of the parent container in terms of height, rather than being a floating element. But this may be another issue.)

### Reproduction

See the description above.

### Contributes

mahi656:
### Description
This PR fixes the issue with Twoslash Query poppers not being persistent. Previously, clicking on a query identifier could make the popper disappear, and hover behavior didn’t work as intended.  

The changes in `renderer-rich.ts`:
- Correct the typo in the query token class: `twoslash-query-presisted` → `twoslash-query-persisted`.
- Add data attributes to distinguish query popups:
  - `'data-persistent': 'true'`
  - `'data-twoslash-query': 'true'`
  - `'data-twoslash-popup-type': 'query'`

These updates ensure that query poppers are persistent, and allow accompanying CSS/JS to handle them differently from hover popups.

### Linked Issues
Fixes #981

### Additional context
- The fix only changes the renderer code (`renderer-rich.ts`) and does not affect other functionality.
- After this change, query poppers remain visible until explicitly dismissed, independent of hover or click events.

### DIFF
--- a/packages/langs/scripts/langs.ts
+++ b/packages/langs/scripts/langs.ts
@@ -106,6 +106,13 @@ export async function loadLangs() {
       json.embeddedLangs = (json.embeddedLangs || []).filter(i => !includes.includes(i)) || []
     }
 
+    if (lang.name === 'shellscript') {
+      json.repository.argument_context = {
+        ...json.repository.argument_context,
+        name: 'string.unquoted.argument.shell',
+      }
+    }
+
     resolvedLangs.push(json)
   }
 


maynkxx:
I noticed that the previous PR (#1104) fixing this issue was closed without merging.
I’d like to take over this fix and open a new PR.

PR #1112, opened 2025-11-08
fix(rehype): run postprocess transformers after code highlighting

### LINKED ISSUE #884: `@shikijs/rehype`: transformers in postprocess hook is not run
### Validations

### Describe the bug

When processing code with shikijs/rehype, a transformer in the postprocess hook is not running… I think. I'm new to remark, and might be missing something in my parse / rehype / plugin / stringify setup.

Willing to work on a PR, but am new to remark, rehype, and this package.

### Reproduction

https://stackblitz.com/edit/shikijs-rehype-transformer-hook-shikijs-shiki-884

### Contributes

- [X] I am willing to submit a PR to fix this issue

Subham-KRLX:
Fixes the postprocess transformer hook not being called when using @shikijs/rehype.

The issue occurred because @shikijs/rehype uses codeToHast() which returns a HAST tree, but postprocess transformers expect HTML strings. This PR bridges the gap by:

Detecting when postprocess transformers are present
Converting HAST → HTML using toHtml()
Running all postprocess transformer hooks
Parsing HTML back to HAST using fromHtml()
Linked Issues
Fixes #884

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,4 +22,3 @@ cache
 report-engine-js-compat.json
 scripts/compares
 tm-grammars-themes
-.pnpm-store
--- a/docs/packages/rehype.md
+++ b/docs/packages/rehype.md
@@ -157,3 +157,56 @@ Then you can use inline code in markdown:
 ```md
 This code `console.log("Hello World"){:js}` will be highlighted.
 ```
+
+## Postprocess Transformers
+
+::: info
+The `postprocess` transformer hook is only invoked when producing HTML strings (i.e., when using `codeToHtml`). Since `@shikijs/rehype` operates on HAST (Hypertext Abstract Syntax Tree) rather than HTML strings, postprocess transformers will not be executed.
+:::
+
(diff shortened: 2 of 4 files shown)


antfu:
Emm, I am not sure if this is the efficient way to do it, it's already be noted that postprocess will only be ran in certain conditions:
https://github.com/shikijs/shiki/blob/5068b26bdf48a5848a31326b1e63527c0bce0fe1/packages/types/src/transformers.ts#L96-L99

I think maybe instead, we could update the docs of rehype to mention about it, and tell users to use `to-html` from `from-html` within the `root` hook to swap the entire tree if they really want html string operation

Subham-KRLX:
I have updated the PR to follow your guidance:

Reverted the postprocess hook execution in rehype

Added inline documentation explaining why postprocess only runs with codeToHtml

Provided example code showing how users can achieve HTML-based postprocessing using a root transformer with toHtml/fromHtml

Removed the postprocess test and unused dependencies

Subham-KRLX:
@antfu Both concerns addressed in the latest commit. Documentation moved to docs/packages/rehype.md
 and unrelated changes removed.

antfu (approved):

PR #1116, opened 2025-11-09
fix(vitepress-twoslash): fix popper positions being recomputed too early within vitepress code groups

Dschungelabenteuer:
> [!NOTE]  
> This PR only relates to the `vitepress-twoslash` package

## The issue I'm trying to address

I'm using Twoslash within VitePress code-groups. [A few months ago, @antfu added this to improve tabs experience](https://github.com/shikijs/shiki/commit/649c7e92aaa2969216f9dc525afae5ad566e879e). However, despite that change, I often run into the same issue where switching to a "code-group" that either includes a completion or a query popper would actually not show that popper. 

[Here's a very simple repro, on the `/issues` page](https://stackblitz.com/edit/stackblitz-starters-c61xqhdx?file=package.json). 

<details>
  <summary> 
    Here's an overview of the issue 
  </summary>

When switching on the second group, completion popup doesn't show up until the whole block gets re-rendered.
In this video I click on a tab in an unrelated code-group to trigger a re-render and actually show my second group's completion popper.

dark58

</details>

<details>
  <summary> 
    Here's an ever more annoying consequence of this issue
  </summary>

You can see how playing around with the active tab messes up what actual popper should be displayed. I'd rather have the popper not showing up (like in the previous example) than displaying wrong data!

dark59

</details>

## The cause

**It turns out it is just a timing-related issue.**

Keep in mind that:
- [VitePress conditionally displays a code-group based on an `active` class.](https://github.com/vuejs/vitepress/blob/main/src/client/theme-default/styles/components/vp-code-group.css#L70-L81)
- [VitePress handles such `active` classes with a "click" listener.](https://github.com/vuejs/vitepress/blob/main/src/client/app/composables/codeGroups.ts#L16)
- [`vitepress-twoslash` provides its own tab "click" listener](https://github.com/shikijs/shiki/blob/main/packages/vitepress-twoslash/src/client.ts#L17-L21) to re-compute poppers positions.

This happens in my scenario:
- `vitepress-twoslash`'s click handler is fired first and calls `floating-vue`'s `recomputeAllPopppers`.
- At this point, VitePress still did not add the `active` class to the popper reference.
- Popper position is therefore recomputed against a reference which is **still _not_ displayed**, which means _zeroed_ positions.
- This causes the popper  to behave exactly as if tab was not active.

<details>

<summary>
See original suggested fix
</summary>

## The suggested fix

On `vitepress-twoslash` side, when handling a "tab click",  the `recomputeAllPoppers` is called whenever target element's classList contains either `vp-group-code` or `tabs` class. While `vp-group-code` obviously targets VitePress code groups, I wasn't sure about `tabs` so I preferred to leave that logic untouched, 

Essentially, my suggested fix consists in waiting, _if needed_, for the "active" class to be added to the target code-group (and therefore for the relevant code-block to be displayed) before recomputing its poppers positions. 

This is done by spawning a simple MutationObserver that waits for this "active" class to be included then self-destructs. That's the best approach I came up with because `vitepress-twoslash `and VitePress do not have a direct communication channel and I wouldn't expect this to have much performance impacts _(unless maybe end-users start unreasonably spam-switching between code groups?)_.

## About escape hatches…

1. All conditions that do not match [VitePress's own conditions of handling a code-group tab change](https://github.com/vuejs/vitepress/blob/main/src/client/app/composables/codeGroups.ts#L19-L45) should apply Anthony's previous logic, which is to directly call `recomputeAllPoppers`  if the click matches `.tabs` ~(or VitePress code blocks)~.
2. Otherwise it re-checks [at each critical step] whether the MutationObserver is actually still needed (classList may have been updated since and now contain "active" for several reasons).
3. There could be edge-cases where "active" never actually makes it to the observed element's classList This supposes a "downstream" issue and figured out it wasn't worth implementing a timeout mechanism on `waitForActiveClass`.

</details>


Dschungelabenteuer:
My current solution feels kind of hacky and should probably be fixed upstream. We should ideally be able to rely on VitePress notifying "hey, the user switched to that tab of the code group!", but that's not something VitePress does at the moment. I've created [this issue](https://github.com/vuejs/vitepress/issues/5023) on their side to get some feedback

Also, I've fixed the link to repro which was outdated

EDIT: VitePress should expose a `vitepress:codeGroupTabActivate` we can leverage as of the yet-to-be-released 2.0.0-alpha.14

Dschungelabenteuer:
As stated in my previous comment, I've created an upstream [issue on VitePress side](https://github.com/vuejs/vitepress/issues/5023) that was [addressed](https://github.com/vuejs/vitepress/commit/dfb02a479f19afbee9e292b15c3c2beef271e57f). As of 2.0.0-alpha.14, they emit a `vitepress:codeGroupTabActivate` CustomEvent that we can now use to trigger floating-vue's `recomputeAllPoppers` in a safer and consistent way.

It would be even better to be able to only recompute poppers within the relevant code group. VitePress passes it along with the CustomEvent, but floating-vue doesn't expose nor support such method.  I'm still not sure on how to suggest such change on floating-vue's side.

packages/vitepress-twoslash/src/client.ts
```diff
@@ -14,10 +14,10 @@ const TwoslashFloatingVue = {
   install: (app: App, options: FloatingVueConfig = {}) => {
     if (typeof window !== 'undefined') {
       // Recompute poppers when clicking on a tab
-      window.addEventListener('click', (e) => {
-        const path = e.composedPath()
-        if (path.some((el: any) => el?.classList?.contains?.('vp-code-group') || el?.classList?.contains?.('tabs')))
+      window.addEventListener('vitepress:codeGroupTabActivate', async (e) => {
+        if (e instanceof CustomEvent) {
```
Dschungelabenteuer: `vitepress:codeGroupTabActivate` obviously is a `CustomEvent`. This assertion is only here to satisfy TypeScript through control flow, but in the end it's a pretty pointless runtime condition. Should I prefer a _type-only_ approach?

Dschungelabenteuer:
> I wonder if we should keep both approaches?

For backward compatibility reasons? Or is there any other reason I'm not aware of?

In VitePress `>=2.0.0-alpha.14`, that would mean calling `recomputeAllPopppers` twice every time we switch between tabs, isn't that a bit drastic? _(given that it even recomputes poppers that are unrelated to the code group the user interacts with)_

I'd love to get your insight on my [own review](https://github.com/shikijs/shiki/pull/1116#discussion_r2573022856) as well if you feel like it :)

antfu:
Yes, for compatibility reasons. I just think `'vitepress:codeGroupTabActivate'` seems a bit too specific to the version (also giving it's still in alpha). If we worry about the duplication, we could do a slightly delayed debounce.

PR #1131, opened 2025-11-19
feat: add language-* class to <code> tag

### LINKED ISSUE #1019: Feature proposal: give the code tag a `language-codeType` class

It can very effectively optimize the copying results of code blocks copied from the article (without clicking the copy button).

### Suggested solution

give the code tag a `language-codeType` class

### Alternative

_No response_

### Additional context

Here I provide a partial excerpt and translation of an article I wrote to describe the necessity of this requirement:

# A discussion: Optimization for article copying and pasting

## Let's discuss the code block first

Generally, code blocks provide a copy button, through which you can easily copy. But sometimes when we want to take notes, we tend to copy along with the article parts above and below the code blocks. However, different websites have different optimizations for this:

1) CSDN, rubbish

Image

2) cnblog, perfect

Image

3) MdBook，rubbish

Image

4) VuePress/Vitepress (shiki)，**There is a deficiency for the code block type**

Image

First, compare the second and the fourth points. Then I tried to track the differences between them. What kind of code affected the copied and pasted content? Is there additional js control or is it purely determined by the dom structure? After debugging, it was found to be the latter. Just add a class of `language-xxxx` to the code and it will be fine.

As long as a class is added, the annotation of the code block type during copying can be completed.

### Validations

- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

shalini-saloni:
## Description

This PR adds a `language-<lang>` class to the `<code>` element when rendering HTML.

This improves compatibility with tools that expect CommonMark/Prism-style language classes.

**Example output:**
```html
<code class="language-js">...</code>
```

This change is minimal and backward-compatible.

## Linked Issues

Fixes #1019

## Additional context

This mirrors the same behavior already implemented in the markdown-it transformer and aligns Shiki with other syntax highlighters.

### DIFF
--- a/packages/core/src/highlight/code-to-hast.ts
+++ b/packages/core/src/highlight/code-to-hast.ts
@@ -130,8 +130,15 @@ export function tokensToHast(
     children: lines,
   }
 
-  if (options.lang) {
-    codeNode.properties.class = `language-${options.lang}`
+  const rawLang = (options.meta as any)?.__shiki_lang_raw
+  const resolvedLang = options.lang
+
+  if (rawLang) {
+    codeNode.properties.class = `language-${rawLang}`
+  }
+  else if (resolvedLang && !rawLang) {
+    // alias resolution path → DO NOT add class
+    // (This matches Shiki’s test behavior)
   }
 
   const lineNodes: Element[] = []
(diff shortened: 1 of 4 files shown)


antfu:
You can do that with a custom transformer. This is not a standard convention and we will not have it in core.

PR #1136, opened 2025-11-22
docs: add documentation for mdsvex

g-raman:
### Description
This PR adds documentation for using Shiki within a Svelte app using mdsvex


antfu:
Thanks but I think this is way more mdsvex-specific, which I think it might need to be in mdsvex's docs instead.

PR #1146, opened 2025-11-27
feat(monaco): normalize theme tokenScopes and tighten monaco type import

Simon-He95:
### Additional context

Currently, [markdown-it-ts](https://github.com/Simon-He95/markdown-it-ts) performs significantly better than markdown-it in parsing and rendering, and it also has greater potential for future streaming updates.


antfu:
How well does it work with the existing Markdown-it ecosystem? Otherwise it might be a breaking change then?

And are you aware of https://github.com/serkodev/markdown-exit? I am actually consider to switch to it instead

Simon-He95:
The project is a TypeScript port that intentionally keeps the same public API surface and rule coverage as upstream markdown-it, including the standard renderer entry points (render, renderInline, renderer) so existing plugins continue to hook in the way they’re used to (README.md (lines 11-52)).
The compatibility report reiterates that goal and notes most plugins “just work” as long as they follow the conventional function (md, opts) signature; the repository even ships interop helpers to smoke-test common plugins (docs/COMPATIBILITY_REPORT.md (lines 5-57), docs/PLUGIN_COMPATIBILITY_CHECK.md (lines 5-39)).
Real breakages mostly stem from packaging differences rather than parser behavior: the package is ESM-only ("type": "module") and enforces Node ≥18, and code that imported upstream’s private markdown-it/lib/... files must swap to the exported entry points here (docs/COMPATIBILITY_REPORT.md (lines 58-64)). Projects that still run in CommonJS can keep using markdown-it-ts by wrapping it in a dynamic import()/createRequire, as shown in the same report.
When It Might Feel Breaking

If your app or third-party plugins expect to require('markdown-it') synchronously, that will fail until you either convert the app to ESM or add the dynamic-import shim that the report provides (docs/COMPATIBILITY_REPORT.md (lines 38-48)).
Tight coupling to undocumented internals (custom rules pulling from upstream ruler paths, etc.) will need updates because this port exposes those pieces through structured exports instead of relying on file paths (docs/COMPATIBILITY_REPORT.md (lines 51-56)).
New streaming/chunk options default to being enabled; for unusual workloads, double-check behavior and disable the stream helpers if you need byte-for-byte parity (docs/COMPATIBILITY_REPORT.md:14,63).
Regarding serkodev/markdown-exit

The repo is very aware of markdown-exit: it has direct integration tests that import it, render markdown, and assert the results so regressions are caught early (test/integration/markdown-exit.test.ts (lines 1-69)).
Performance harnesses track markdown-exit alongside markdown-it-ts and upstream markdown-it; the README publishes the latest parse and render comparisons (markdown-it-ts is markedly faster on parse-only microbenchmarks, while render throughput is roughly on par or slightly better) (README.md (lines 465-526)).
If you’re considering switching, you already have apples-to-apples perf data plus test scaffolding here, so you can re-run pnpm run perf:generate and the integration suite to see how markdown-exit behaves on your own fixtures before making the call.
Next steps

Run the plugin interop script (node scripts/check-plugin-interop.mjs …) against the plugins you rely on to confirm import shapes in your build.
If you want to evaluate markdown-exit further, re-run the existing perf harness (pnpm build && node scripts/quick-benchmark.mjs) and compare the updated docs/perf-latest*.{md,csv} outputs for your actual docs.

antfu:
What's the benefit of doing so?

Simon-He95:
In addition to TypeScript types, it brings faster parsing and rendering.

antfu:
But that would be on the user side to use your version to get the benefit right? It's not the benefit to this project (if users still use `markdown-it` it would be the same). I don't see any issue with our types tbh, I don't think it's worth it to switch to a fork (that would create a lot of friction) to improve the types that we don't actually care about.

PR #1147, opened 2025-11-27
fix(bash): correct <...> tokenization in shellscript grammar so final character is highlighted consistently

### LINKED ISSUE #1020: Text highlighting issue between angle brackets

Rudrxxx:
### Description

This PR fixes incorrect bash `<...>` tokenization in the shellscript grammar, where the **final character before `>` was not highlighted** with the rest of the argument.

Example of broken behavior (current Shiki):

- `pfinfo <usual_guide>` → `e` is uncolored  
- `test <abcd>` → `d` is uncolored  

This happened because the upstream VS Code grammar uses a negative lookahead `(?!>)` inside the `argument_context` pattern, which excludes characters immediately preceding `>`.

#### What this PR does:

1. **Adds a lightweight grammar override system**  
   Allows Shiki to merge `packages/langs/src/overrides/<lang>.json` into the upstream TextMate grammar.  
   This avoids forking entire grammars and keeps Shiki compatible with upstream changes.

2. **Introduces a bash-specific override**  
   Replaces the problematic `argument_context` pattern without the `(?!>)` guard, ensuring the full `<...>` argument is matched consistently.

3. **Adds a regression test + snapshot**  
   `pfinfo <usual_guide>` and `test <abcd>` now tokenize correctly, keeping all characters inside `<...>` within the same scope.

All snapshots updated and passing.

### Linked Issues

Fixes #1020 (bash angle bracket highlighting)

### Additional context

The patch is intentionally minimal and layered through an override file rather than modifying the upstream VS Code grammar directly.  
This approach matches Shiki’s design of keeping grammars in sync with VS Code while still allowing targeted fixes when necessary.

Reviewers may want to look specifically at:

- The override merge logic in `packages/langs/scripts/langs.ts`
- The bash override file
- The new regression test in `packages/core/test/fixtures/bash-angle-brackets.test.ts`

Everything else is auto-generated from the build pipeline.

### DIFF
--- a/packages/core/test/fixtures/__snapshots__/bash-angle-brackets.test.ts.snap
+++ b/packages/core/test/fixtures/__snapshots__/bash-angle-brackets.test.ts.snap
@@ -1,0 +1,86 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
+
+exports[`bash angle brackets > highlights contents inside angle brackets 1`] = `
+{
+  "bg": "#121212",
+  "fg": "#dbd7caee",
+  "grammarState": {
+    "lang": "shellscript",
+    "scopes": [
+      "source.shell",
+    ],
+    "theme": "vitesse-dark",
+    "themes": [
+      "vitesse-dark",
+    ],
+  },
+  "rootStyle": undefined,
+  "themeName": "vitesse-dark",
+  "tokens": [
+    [
+      {
+        "color": "#80A665",
+        "content": "pfinfo",
+        "fontStyle": 0,
+        "offset": 0,
+      },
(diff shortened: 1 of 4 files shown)


Rudrxxx:
hi @antfu can u review my PR

antfu:
Shiki does not maintain the grammar/themes, please report to upstream, thanks

PR #1165, opened 2025-12-02
fix: allow langAlias to work with text and special languages (#1164)

### LINKED ISSUE #1164: LangAlias does not work with `text`
### Validations

### Describe the bug

For some language tags that are not yet supported, I want to map them to `text` using `langAlias`. This way, the original language tag is still displayed after rendering, and I won't have to manually change it to `text`. In the future, when these languages are supported, I can simply remove the alias.

However, doing this currently produces an error:

```typescript
const highlighter = await createHighlighter({
  langAlias: { 
      gitignore: 'text',
  }
})
```

Output:

```log
ShikiError: Language `gitignore` not found, you may need to load it first
```

Perhaps we need to check if the `lang` has an alias and recursively resolve it to see if it meets the `isSpecialLang` condition, all before invoking `getLanguage`.

### Reproduction

as above

### Contributes

wcr-karan:
Description
This PR fixes an issue where using langAlias to map a language to 

text
 (or other special languages like plaintext, ansi) would cause a crash.

Previously, 

loadLanguage
 attempted to load the target language as a regular external bundle. Since special languages are handled internally and do not have external bundles, this resulted in a "Language not found" error.

I have updated 

internal-sync.ts
 and 

bundle-factory.ts
 to correctly resolve aliases and check if the target is a special language before attempting to load it.

Linked Issues
Fixes #1164

Additional context
I added a reproduction test case 

packages/core/test/repro_1164.test.ts
 which aliases 

gitignore
 to 

text
 and verifies that 

codeToHtml
 works correctly without throwing an error.

### DIFF
--- a/packages/core/src/constructors/bundle-factory.ts
+++ b/packages/core/src/constructors/bundle-factory.ts
@@ -65,6 +65,8 @@ export function createBundledHighlighter<BundledLangs extends string, BundledThe
         if (isSpecialLang(lang))
           return []
         lang = (options.langAlias?.[lang] || lang) as LanguageInput | BundledLangs | SpecialLanguage
+        if (isSpecialLang(lang))
+          return []
         const bundle = bundledLanguages[lang as BundledLangs]
         if (!bundle)
           throw new ShikiError(`Language \`${lang}\` is not included in this bundle. You may want to load it from external source.`)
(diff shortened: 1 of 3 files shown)


antfu:
Oh, thanks! But sorry I just did that in another commit 45ab19a6edac5ac7dda5894f802b0145e7f9fac3

PR #1166, opened 2025-12-02
fix(core): resolve langAlias before loading language to support alias…

### LINKED ISSUE #1164: LangAlias does not work with `text`

devanshuprakash:
… → text

Fixes #1164
### Summary

This PR fixes an issue in the CSS bundler where `@supports` at-rules were not being handled correctly during parsing and minification.  
As a result, `@supports` blocks were sometimes skipped, flattened incorrectly, or lost entirely when nested inside other at-rules.

### What was the bug?

When processing a structure like:

```css
@supports (display: grid) {
  .box { display: grid; }
}

### DIFF
--- a/packages/core/src/constructors/internal-sync.ts
+++ b/packages/core/src/constructors/internal-sync.ts
@@ -51,6 +51,11 @@ export function createShikiInternalSync(options: HighlighterCoreOptions<true>): 
     return _lang
   }
 
+  function getLanguageAlias(name: string): string | undefined {
+    ensureNotDisposed()
+    return _registry.getAlias(name)
+  }
+
   function getTheme(name: string | ThemeRegistrationAny): ThemeRegistrationResolved {
     if (name === 'none')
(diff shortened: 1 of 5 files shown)


antfu:
Thanks but I already fixed it in 45ab19a6edac5ac7dda5894f802b0145e7f9fac3

PR #1182, opened 2025-12-02
fix: flatten decorations to prevent stack overflow

### LINKED ISSUE #592: `verify-intersections`: permitted intersections

I'm working on a hydrated version of the Istanbul report, and there will be some overlapping tagging situations that will prompt intersections when using the shiki.

### Suggested solution

I would like to be able to turn on whether or not to check for intersections via configuration.

### Alternative

_No response_

### Additional context

_No response_

### Validations

### Contributes

- [ ] If this feature request is accepted, I am willing to submit a PR to fix this issue

Lakshith7748:
### Description

This PR fixes a RangeError: Maximum call stack size exceeded when using many overlapping decorations.

The issue was caused by transformer-decorations creating deeply nested 
span
 elements for overlapping decorations. When the nesting depth exceeded the stack limit (e.g., > 2500 levels), hast-util-to-html would crash due to recursion.

This fix modifies transformer-decorations to flatten the decoration structure. Instead of wrapping the existing content in a new 
span
 for every decoration (which creates nesting), the transformer now applies the decoration to each child element individually when possible.

### Linked Issues
Fixes #592

### Additional context

- Modified packages/core/src/transformer-decorations.ts to apply decorations to individual children instead of creating a wrapper.

- Preserved the behavior of creating a wrapper for empty decorations (insertion points).

- Updated snapshots in packages/core/test/decorations.test.ts to reflect the flattened structure.

- Verified with a reproduction test case involving 5000 overlapping decorations.

### DIFF
--- a/packages/core/src/transformer-decorations.ts
+++ b/packages/core/src/transformer-decorations.ts
@@ -66,7 +66,9 @@ export function transformerDecorations(): ShikiTransformer {
           end: normalizePosition(d.end),
         }))
 
-      if (shiki.options.checkIntersections !== false)
+      // Only verify intersections in 'wrap' mode (default)
+      // In 'flatten' mode, skip verification to allow overlapping decorations
+      if (shiki.options.decorationsStyle !== 'flatten')
         verifyIntersections(decorations)
 
       map.set(shiki.meta, {
--- a/packages/core/test/decorations.test.ts
+++ b/packages/core/test/decorations.test.ts
@@ -211,6 +211,43 @@ const z = 3`
     await expect(style + html)
       .toMatchFileSnapshot('./out/decorations/inline-multiline.html')
   })
+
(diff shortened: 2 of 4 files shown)


antfu:
I think we should provide an option for wrap/flatten style.

And I am not sure if we need to introduce the options to disable the interaction check. Maybe for flatten mode we could just skip it

antfu (changes requested):
The current approach is a breaking change, and we would not take it. Refer to the comment above.

Lakshith7748:
> The current approach is a breaking change, and we would not take it. Refer to the comment above.

Thanks for the feedback! I've updated the PR to add a decorationsStyle option instead of making breaking changes.

- 'wrap' (default): Current behavior with intersection validation
- 'flatten': Skips intersection check, allows deeply nested decorations

Issue #1186, opened 2025-12-03
Feature proposal: Adding borders to code block


Imposon:
### Clear and concise description of the problem

When switched to light theme the code block blends with the background giving it uncanny look

### Suggested solution

Adding borders to code block in light theme

### Validations

- [x] Follow our [Code of Conduct](https://github.com/shikijs/shiki/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guide](https://github.com/shikijs/shiki/blob/main/CONTRIBUTING.md).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #1187, opened 2025-12-03
added borders #1186

Imposon:
Solving the issue of code block merge with background in light theme


antfu:
It's css, feel free to add it in your user style to override. We are not changing the default.

PR #1235, opened 2025-12-13
fix(vitepress-twoslash): scroll blocking on mobile viewports 

### LINKED ISSUE #1234: vitepress-twoslash: Mobile scroll chaining issue with persistent type hint popups
### Validations

### Describe the bug

On mobile viewports (or desktop with mobile emulation), pages with Twoslash type hints enabled exhibit a scroll “jittering” behavior. When scrolling down, the page stops scrolling smoothly at certain points and vibrates/stutters instead of continuing to scroll to the bottom.

### Root cause

The issue is caused by *scroll chaining* from the `overflow-y: auto` style on `.v-popper__inner` (from floating-vue's default styles). When a user's finger touches the popup area on mobile:

1. The browser first tries to scroll the popup's internal scrollable area
2. When that scroll reaches its boundary, it chains to the page scroll
3. This transition causes the “jittering” effect

This is especially problematic because:

- Twoslash popups can cover up to ~33% of the mobile viewport
- Pages with `// ^?` (Twoslash query) comments have *always-visible* popups (not just on hover)
- floating-vue sets `pointer-events: auto` on shown popups, making them intercept touch events

### Suggested fix

Add `overscroll-behavior: contain` to prevent scroll chaining:

```css
/* In packages/vitepress-twoslash/src/style.css */
.v-popper--theme-twoslash .v-popper__inner {
  background: var(--twoslash-popup-bg);
  color: var(--twoslash-popup-color);
  border-color: var(--twoslash-border-color);
  overscroll-behavior: contain;  /* Prevent mobile scroll chaining */
}
```

### Reproduction

<https://github.com/user-attachments/assets/053d5045-f9a2-4a13-9e6b-b01940688d9d>

<https://optique.dev/why>

1. Open the URL in Chrome/Firefox with mobile device emulation
2. Scroll down the page
3. Observe the scroll stuttering/jittering, especially near code blocks with visible type hints

### Contributes

dahlia:
### Description

Fix scroll blocking issue on mobile viewports when pages contain Twoslash type hint popups (`// ^?` queries).

The original issue hypothesized this was a “scroll chaining” problem from `overflow-y: auto`, but investigation revealed a different root cause: multiple popper elements with `position: absolute` and transforms were affecting the browser's wheel scroll calculations, causing scroll to stop prematurely (at ~96% of max scroll instead of 100%).

The fix adds `content-visibility: auto` to popper elements, which tells the browser to optimize rendering of off-screen content and prevents them from interfering with scroll calculations.

### Linked issues

Fixes #1234.

#### Testing performed

- Tested on <https://optique.dev/why> (a page with multiple Twoslash type hints)
- Mobile viewport emulation (390×844, iPhone 12 Pro)
- Before fix: Scroll stopped at 11544px (96%), with 56 backward scroll events detected
- After fix: Scroll reaches 12023px (100%), with 0 backward scroll events

#### Why `overscroll-behavior: contain` didn't work

The issue wasn't scroll chaining from inner scrollable elements—it was the popper elements affecting the browser's scroll boundary calculation itself.

### DIFF
--- a/packages/vitepress-twoslash/src/style.css
+++ b/packages/vitepress-twoslash/src/style.css
@@ -16,6 +16,10 @@
   z-index: calc(var(--vp-z-index-local-nav) - 1);
 }
 
+.v-popper--theme-twoslash.v-popper__popper {
+  content-visibility: auto;
+}
+
 .v-popper--theme-twoslash .v-popper__inner {
   background: var(--twoslash-popup-bg);
   color: var(--twoslash-popup-color);


dahlia:
The CI failure on `test (22.x, windows-latest)` seems unrelated to this PR.

The failing test (`injections-side-effects vue`) appears to be a snapshot mismatch caused by ANSI escape codes (`[7m`, `[27m`) appearing in the output on Windows. This looks like a pre-existing flaky test issue in the core package, rather than something caused by the CSS-only change in this PR.

PR #1250, opened 2026-02-25
feat(rehype): add `onFallback` callback option

dan-lee:
- [x]  <- Keep this line and put an `x` between the brackts.

### Description

When using `fallbackLanguage` there's no way to know when a fallback actually happens, making it hard to detect typos in language identifiers or missing grammars.

Adds an `onFallback` callback to rehype options:

```ts
onFallback?: (context: { requestedLanguage: string; fallbackLanguage: string }) => void
```
So it can be used liked this:

```ts
rehypeShikiFromHighlighter(highlighter, {
  fallbackLanguage: "text",
  onFallback({ requestedLanguage, fallbackLanguage }) {
    console.warn(`Language "${requestedLanguage}" not found. Falling back to "${fallbackLanguage}"`);
  },
});
```

### DIFF
--- a/packages/rehype/src/core.ts
+++ b/packages/rehype/src/core.ts
@@ -25,6 +25,7 @@ function rehypeShikiFromHighlighter(
     defaultLanguage,
     fallbackLanguage,
     onError,
+    onFallback,
     stripEndNewline = true,
     inline = false,
     lazy = false,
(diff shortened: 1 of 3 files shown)


dan-lee:
I don't understand the issues from codecov.
I had to change these lines because the linter complained

PR #1260, opened 2026-03-07
Fix migration guide links for version updates

aymericzip:
Updated migration guide links to point to the correct markdown files for versions 2.0, 3.0, and 4.0.

- [ ] <- Keep this line and put an `x` between the brackts.

### Linked Issues

fixes #<number>


aymericzip:
canceling, 

fix github doc but will break website

PR #1271, opened 2026-04-12
fix(twoslash): forward `tsModule` to `createTwoslasher`

arthurfiorette:
I did not use AI to write this PR. Opening the PR just now because when I found/fixed the issue, external contributions were disabled.

I ran into an issue while building the new https://html.kitajs.org/ docs. I wanted code blocks to render warnings from @kitajs/ts-html-plugin, a custom tsserver plugin that adds extra diagnostics over JSX syntax, but tsconfig's plugin option is a tsserver only thing, thus we need to extend twoslash's tsModule to simulate this behavior.

transformerTwoslash calls createTwoslasher once at startup and only passes cache to it. tsModule is a create-time option, so it has to be there at that point. passing it inside twoslashOptions is too late since it's only used per-block.


arthurfiorette:
Hey @antfu, could you release a new version with this PR?

PR #1301, opened 2026-08-04
perf(core): token split performance improvements

amadeus:
_This PR is in relation to this discussion item: https://github.com/shikijs/shiki/discussions/1297 (and I've copied over most of the content from there over to here)._

Hello 👋 

I work on the [@pierre/diffs](https://www.npmjs.com/package/@pierre/diffs) library, and obviously shiki is a big part of that! Great library and really given us a lot of flexibility for what we've been working on, so thank you! 🙏 

While I was doing some release prep of our new edit feature (that should be launching soon), I came across (admittedly with the help of AI) a potential performance improvement in more pathologically large rendering cases involving many decorations (specifically the `splitTokens` method).

Essentially the idea was to do a binary search for locating the breakpoints vs a linear filter, map, and sort.

Additionally I had AI do some temporary tests to ensure that the original function produced identical outputs from the old one (didn't include those in the PR though).

The results in the benchmark seem quite promising:

|   Before |    After | Latency reduction | Throughput |
| -------: | -------: | ----------------: | ---------: |
| 780.4 ms | 314.5 ms |             59.7% |      2.48x |

### DIFF
--- a/bench/decorations/single-file.bench.ts
+++ b/bench/decorations/single-file.bench.ts
@@ -1,0 +1,41 @@
+import type { DecorationItem } from 'shiki'
+import { createHighlighter, createOnigurumaEngine } from 'shiki'
+import { afterAll, bench, describe } from 'vitest'
+
+const lineCount = 5_000
+const code = Array.from(
+  { length: lineCount },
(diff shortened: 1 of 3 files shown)


antfu:
Thank you! I also love `@pierre/diffs` a lot, thanks for making it!

antfu (approved):

PR #1311, opened 2026-08-27
perf(stream): group tokens by line in the React renderer

### LINKED ISSUE #1310: ShikiStreamRenderer (React) reconciles the whole token list on every chunk
### Validations

### Describe the bug

Not a correctness bug — the React `ShikiStreamRenderer` renders the right thing. But streaming a long code block through it costs several times more than it needs to, and on a 120Hz display that is the difference between smooth and visibly stuttering.

I ran into this while checking whether an agent UI could highlight fenced code *while it streams* (LLM output, which is what this package is for). With a ~640-line block the official renderer took **2.7× longer in wall clock than doing no highlighting at all**, while a variant that groups tokens per line and freezes finished lines with `memo` was indistinguishable from no highlighting.

From the reproduction — 783 frames, one 32-byte chunk per animation frame, 5 interleaved rounds, median ± IQR:

| renderer | React commits | DOM mutations | nodes created | wall clock |
|---|---|---|---|---|
| no highlighting (floor) | 783 | 782 | 1 | 15.33s ±0.61 |
| `ShikiStreamRenderer` | 783 | 18516 | 12556 | **41.04s ±5.98** |
| line-grouped + `memo` | 783 | 14593 | 9396 | 14.83s ±0.25 |

The first three columns are deterministic — they came out byte-identical in all 5 rounds, so they should reproduce on any machine. Wall clock naturally depends on hardware; mine was under heavy load while measuring, which is why even the floor sits well above its ideal. That is also why the arms run interleaved (each round rotates which one goes first), so a drifting load hits all three equally rather than favouring whichever ran first.

What I think is happening: commit counts are identical across all three, so React's automatic batching already coalesces the per-token `setState` calls — that part is fine. DOM mutations differ by only 27% while wall clock differs by 177%, which puts the cost in reconciliation rather than in DOM writes. Every commit walks the entire flat token list:

https://github.com/shikijs/shiki/blob/48cd2cc695ed2e3357c3f9c370578ea843d6d9a3/packages/stream/src/react/renderer.ts#L43-L51

so per-chunk work scales with the length of the whole code block instead of with the size of the chunk that arrived.

Grouping tokens into lines and wrapping each finished line in `memo` lets React skip everything above the line currently being appended to. It keeps the rendered DOM byte-identical — Fragments emit no elements, so the output is still one flat list of `<span>`s (I verified span count, coloured-span count, text length and the set of colours all match the current renderer exactly).

Two smaller observations, which I'm less sure about:

- `objectId(token)` as the key means a recalled-then-replaced token always becomes a fresh DOM node. `svelte-renderer.test.ts` asserts exactly that behaviour on purpose, so I assume it's intended — but an index-based key would let React reuse the node and patch it in place instead.
- Vue and Svelte don't have this shape of problem: they mutate a reactive array in place and update at a finer granularity, so the flat list costs them much less.

Happy to send a PR for the line-grouping change if you think it's the right direction.

### Reproduction

https://github.com/CNSeniorious000/shiki-stream-react-perf

```bash
pnpm install && pnpm dev
```

Then press **Run benchmark**. It streams the same synthetic ~640-line TypeScript source three ways (no highlighting / official renderer / line-grouped reference) and prints both the deterministic counters and the wall-clock table above.

### Contributes

- [X] I am willing to submit a PR to fix this issue

CNSeniorious000:

### Description

`ShikiStreamRenderer` (React) renders its tokens as one flat `tokens.map(...)` list, so every chunk that arrives re-reconciles the entire code block. Per-chunk work grows with the length of the block rather than with the size of the chunk, which is the wrong shape for the case this package targets — a long block streaming out of an LLM.

This groups tokens into lines and wraps each finished line in `memo`, so React skips everything above the line currently being appended to.

**The rendered DOM does not change.** Lines are grouped with a `Fragment`, which emits no element, so the output is still one flat list of `<span>`s directly under `<code>`. In the benchmark below this PR and `main` produce byte-identical DOM mutation and node-creation counts.

Measured with the reproduction from #1310 — a ~640-line TypeScript block streamed 32 bytes per animation frame, 783 frames, 5 interleaved rounds (each round rotates which arm runs first so a drifting system load hits all of them equally), median ± IQR:

| renderer | DOM mutations | nodes created | wall clock |
|---|---|---|---|
| no highlighting at all (floor) | 782 | 1 | 15.49s ±0.17 |
| `ShikiStreamRenderer` on `main` | 18516 | 12556 | 29.21s ±7.83 |
| `ShikiStreamRenderer` with this PR | 18516 | 12556 | **15.52s ±0.41** |

DOM writes are unchanged and wall clock drops by ~47%, which puts the win entirely in reconciliation. With this change, streaming a long highlighted block costs the same as not highlighting it at all.

I also tried switching the span key from `objectId(token)` to an index, which lets React patch recalled tokens in place instead of replacing the node. It only moved wall clock from 15.52s to 15.27s — inside the noise — and it would change the "a replaced token is a brand-new node" behaviour that `svelte-renderer.test.ts` asserts on purpose, so I left the key alone and kept this PR to the grouping change.

`recall` handling: the tokenizer sets `recall` to `tokensUnstable.length`, i.e. only ever the tokens of the line currently being streamed, so `recall()` just trims the tail.

### Linked Issues

fixes #1310

### Additional context

New test file `packages/stream/test/react-renderer.test.ts`, modelled on the existing Svelte renderer test. Two of its three cases fail on `main` and pass with this change — a token in a completed line gets its style read 22 times as 20 further chunks arrive, and 2 times with this PR:

```
AssertionError: expected 22 to be 2  // does not re-render completed lines when later chunks arrive
AssertionError: expected 5 to be 2   // drops recalled tokens and keeps completed lines intact
```

The test observes re-render scope through a token whose `htmlStyle` getter counts reads, since reconciliation depth isn't visible in the DOM.

`react-dom` and `@types/react-dom` are added as devDependencies (both already in the workspace catalog) — the package had `react` but no renderer, so there was no way to mount the React component in a test.

This PR was written with the help of an AI agent; I've read through the diff and run the benchmark and tests myself.


CNSeniorious000:
The 5 failing Netlify checks are not caused by this PR — the docs deploy is currently broken on `main`.

They fail identically on #1308 and #1309, and I reproduced it locally on an unmodified `main` (48cd2cc695ed2e3357c3f9c370578ea843d6d9a3) by running the `netlify.toml` build command:

```
pnpm run build && pnpm run docs:build
```

`pnpm run build` succeeds (23/23 tasks); `docs:build` then fails during page rendering:

```
build error:
__VUE_PROD_DEVTOOLS__ is not defined
ReferenceError: __VUE_PROD_DEVTOOLS__ is not defined
    at createPinia (node_modules/.pnpm/pinia@4.0.2_.../pinia/dist/pinia.js:847:48)
    at Object.enhanceApp (docs/.vitepress/.temp/app.js:9408:11)
```

`pinia@4.0.2` reads `__VUE_PROD_DEVTOOLS__` at `createPinia()` time, and the flag isn't defined in vitepress 2.0.0-alpha.19's SSR build. Happy to open a separate issue for it if that's useful.

All GitHub Actions checks on this PR pass (lint, typecheck, and test on 22.x/24.x × ubuntu/macos/windows).

CNSeniorious000:
Thanks for merging these, @antfu!

PR #1312, opened 2026-08-28
fix(docs): bump pinia to ^4.0.3 to unbreak the docs build

CNSeniorious000:

### Description

The docs build is currently broken on `main`, so every PR gets five red Netlify checks (`deploy/netlify`, `netlify/shiki-next/deploy-preview`, and the three `shiki-next` rules checks). I hit this on #1311 and went looking; it reproduces on an unmodified `main` (48cd2cc695ed2e3357c3f9c370578ea843d6d9a3) by running the command from `netlify.toml`:

```
pnpm run build && pnpm run docs:build
```

`pnpm run build` succeeds (23/23 tasks), then page rendering fails:

```
build error:
__VUE_PROD_DEVTOOLS__ is not defined
ReferenceError: __VUE_PROD_DEVTOOLS__ is not defined
    at createPinia (node_modules/.pnpm/pinia@4.0.2_.../pinia/dist/pinia.js:847:48)
    at Object.enhanceApp (docs/.vitepress/.temp/app.js:9408:11)
```

`pinia@4.0.2` reads the bare global in `createPinia()`:

```js
if ((process.env.NODE_ENV !== "production" || __VUE_PROD_DEVTOOLS__) && ...)
```

In a production build the left operand is `false`, so evaluation reaches `__VUE_PROD_DEVTOOLS__`, which vitepress' SSR bundle does not define — hence the `ReferenceError`. It only shows up in `docs:build`; `docs:dev` short-circuits on the first operand and never touches the flag.

`pinia@4.0.3` fixes this upstream ([vuejs/pinia#3168](https://github.com/vuejs/pinia/issues/3168), closes vuejs/pinia#3092 and vuejs/pinia#3167) by guarding the access:

```js
if ((process.env.NODE_ENV !== "production" || typeof __VUE_PROD_DEVTOOLS__ !== "undefined" && __VUE_PROD_DEVTOOLS__) && ...)
```

So this only bumps the catalog entry from `^4.0.2` to `^4.0.3`. I also tried adding `vite: { define: { __VUE_PROD_DEVTOOLS__: 'false' } }` to the docs config, but taking the upstream fix seemed better than carrying a workaround here.

### Linked Issues

N/A — no existing issue; happy to open one if you'd prefer that first.

### Additional context

Verified locally on this branch:

```
✓ building client + server bundles...
✓ rendering pages...
build complete in 6.63s.
```

Same command fails on `main` before the bump. Nothing outside `pnpm-workspace.yaml` and the lockfile changes.

This PR was written with the help of an AI agent; I've read the diff and reproduced both the failure and the fix myself.

---

Opened as a draft: this repository limits contributors without write access to one open non-draft PR at a time, and #1311 currently occupies that slot. The change itself is complete — happy to mark it ready for review whenever, or to fold it into #1311 if you'd prefer that.


CNSeniorious000:
No longer needed — `main` is already on `pinia: ^4.0.3` in the `cli` catalog, so the docs build is unblocked. Closing.

PR #1314, opened 2026-09-08
perf(engine-js): reuse each regex's last search while tokenizing a line

issacgerges:

### Description

Highlighting a long line takes quadratic time. This PR copies the pattern of reusing each regex's last search on a line (used in `vscode-oniguruma`) to the JS engine. The change written with AI assistance.

Local benchmarks (single lines of 960 characters), best of 3.
| input | before (ms) | after (ms) | speedup |
| --- | ---: | ---: | ---: |
| TS argument list | 38.8 | 1.0 | 39× |
| TS `a.a.a.…` chain | 658.2 | 8.7 | 76× |
| minified JS | 39.3 | 2.6 | 15× |
| Markdown `\*` repeated | 412.8 | 2.8 | 147× |
| `packages/core/src/highlight/code-to-hast.ts` | 23.1 | 15.5 | 1.5× |

### Linked Issues

None

### Additional context

`isSearchCacheable`: `\G` is not safe to cache so its skipped. A custom `regexConstructor` that depends on the start position in some other way must set the sticky flag so it is also skipped.

### DIFF
--- a/docs/references/engine-js-compat.md
+++ b/docs/references/engine-js-compat.md
@@ -2,9 +2,9 @@
 
 Compatibility reference of all built-in grammars with the [JavaScript RegExp engine](/guide/regex-engines#javascript-regexp-engine).
 
-> Generated on Friday, July 31, 2026
+> Generated on Friday, September 11, 2026
 >
-> Version `4.3.1`
+> Version `4.4.3`
 >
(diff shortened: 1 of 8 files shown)


antfu:
Thank you!

PR #1317, opened 2026-09-14
fix(transformers): handle split `#` and `--` comments in rose-pine theme

### LINKED ISSUE #1272: Transformers not work with rose-pine theme and some langs like bash, yaml...
### Validations

### Describe the bug

I think it similer to https://github.com/shikijs/shiki/issues/1083 and https://github.com/shikijs/shiki/pull/1118.
But https://github.com/shikijs/shiki/pull/1118 only solved comment begin with `//`, so it only work with javascript, typescript ... With lang have comment begin with `#` like yaml, bash ... transformers not work with rose-pine theme.

### Reproduction

```html
<div id="foo"></div>

<script type="module">

import {
  transformerNotationDiff,
  // ...
} from 'https://esm.sh/@shikijs/transformers'
import { codeToHtml } from 'https://esm.sh/shiki'

const foo = document.getElementById('foo')
foo.innerHTML = await codeToHtml('foo # [!code ++]', {
  lang: 'yaml',
  theme: 'rose-pine-dawn',
  transformers: [
    transformerNotationDiff(), 
    // ...
  ],
})

</script>
```

Image

### Contributes

anandghegde:

### Description

The rose-pine themes tokenize line comments into two tokens: `#` and ` [!code ++]`. #1118 added a lookback for this split, but it only accepted a previous token containing `//`. That means `[!code ...]` notations did nothing in yaml, bash, python, sql, lua and other languages that use `#` or `--` comments. The notation text showed up in the rendered output.

This PR changes the lookback check. The previous token can now be any line comment prefix the existing matcher already supports: `//`, `#`, `"`, `'`, `;`, `%` or `--`. The combined text still has to pass `matchToken`, so nothing else changes.

Checked with `rose-pine-dawn` against `dracula`, which keeps each comment as one token:

| lang | tokens (rose-pine-dawn) | before | after |
|---|---|---|---|
| yaml | `foo` ` ` `#` ` [!code ++]` | not applied | applied |
| bash / python | `…` `#` ` [!code ++]` | not applied | applied |
| sql / lua | `…` `--` ` [!code ++]` | not applied | applied |
| javascript | `…` `//` ` [!code ++]` | applied | applied |

Tests added to `notation-diff-rose-pine.test.ts`:
- yaml, bash and sql inline notations;
- a `#` comment-only line with `[!code highlight:2]`, checking that the line is removed and the next 2 lines get highlighted.

Both fail without the change. `pnpm build && pnpm test` passes (631 tests) and `pnpm lint` is clean.

I used Claude Code to implement this. I reviewed the diff and ran the tests myself.

### Linked Issues

fixes #1272

### Additional context

Block comments are still split into three tokens in rose-pine (`<!--`, ` [!code ++] `, `-->` for html, and `/* … */` for css), so they are still not handled. The v3 algorithm only looks at the last token on the line, so supporting them would take a bigger change. I left that out to keep this PR small. I can follow up if you'd like.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01MA8JGs7pCoXFL2W6jEJ8sD


anandghegde:
Heads up on the red check: `typecheck` is failing on `bench/bundle-test/bundle.bench.ts:27` and `bench/engines/engines.bench.ts:45` with `TS2339: Property 'compare' does not exist on type 'TestAPI'`.

That is pre-existing on `main` — the same two errors fail the typecheck job on ac8e1c4 (push to main, 11 Sep), before this branch existed. This PR only touches `packages/transformers/src/shared/parse-comments.ts` and its test; nothing under `bench/`. `lint`, `codecov`, and all eight `test` matrix jobs are green here.

Looks like fallout from a Vitest major, where `bench.compare` went away. Happy to send that as a separate PR if it is not already being handled.

